PR #17, opened 2016-11-07
Issue12

### LINKED ISSUE #12: Rewrite all accumulation functions
They're all special cases of `accumulateWith`

turion:
Fixes #12. Compiles. If you're unhappy to see the variant with `feedback` go, then comment and I'll rewrite `accumulateWith` in terms of `feedback`.

### DIFF
--- a/src/Data/MonadicStreamFunction/Util.hs
+++ b/src/Data/MonadicStreamFunction/Util.hs
@@ -18,22 +18,22 @@ type MSink   m a = MSF m a ()
 
 -- * Stateful accumulation
 
+accumulateWith :: Monad m => (a -> s -> s) -> s -> MSF m a s
+accumulateWith f s = MSF $ \a -> let s' = f a s in return (s', accumulateWith f s')
+
+-- ** Accumulation for monoids
+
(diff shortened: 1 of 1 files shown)


ivanperez-keera:
I'd say `feedback` seems more fundamental than `accumulateWith`. If so, then it would be best to implement `accumulateWith` with `feedback`, which seems trivially possible.

ivanperez-keera:
Would you mind renaming this branch as develop_issue12, and possibly develop_some_significant_name, so that it's clear by the name what the branch is about, and how it is related to other branches?

turion:
Ok, I'll rewrite `accumulateWith` with feedback, it makes sense. Can we expect performance problems from this, or should I inline it?

I'll happily rename the branch to develop_issue12_rewrite_accumulation_functions, but the only way I know is by creating a new branch and deleting the old one including the pull request... not sure how to do this cleverly.

ivanperez-keera:
There will be a small (linear) overhead from using `arr`. I would not inline anything for now. I'd avoid optimisations for now.

Renaming git branch: Rename locally, remove remotely, push new (old branch) remotely. Yes, pull request could be lost.

PR #30, opened 2017-10-08
Generalised 'once'

turion:
This way, the output of `once` can be used.


turion:
Yes, that's an improvement.

PR #31, opened 2017-10-08
Dev concat s

turion:
Added utility to concatenate an MStream of lists

### DIFF
--- a/dunai.cabal
+++ b/dunai.cabal
@@ -63,6 +63,7 @@ library
                      Data.MonadicStreamFunction.ArrowChoice
                      Data.MonadicStreamFunction.ArrowLoop
                      Data.MonadicStreamFunction.ArrowPlus
+                     Data.MonadicStreamFunction.Async
                      Data.MonadicStreamFunction.Instances
                      Data.MonadicStreamFunction.Instances.Num
                      Data.MonadicStreamFunction.Instances.VectorSpace
--- a/src/Data/MonadicStreamFunction/Async.hs
+++ b/src/Data/MonadicStreamFunction/Async.hs
@@ -35,6 +35,28 @@ Enter a list of Ints:
 2
 Enter a list of Ints:
 ...
+
+Beware that @concatS msf@ becomes unproductive when @msf@ starts outputting empty lists forever.
+This is ok:
+
+>>> let boolToList b = if b then ["Yes"] else []
+>>> let everyOddEmpty = count >>> arr (even >>> boolToList)
+>>> reactimate $ concatS everyOddEmpty >>> arrM print
+"Yes"
+"Yes"
+"Yes"
+"Yes"
+"Yes"
+...
+
+But this will be caught in a loop:
+
+>>> let after3Empty = count >>> arr ((<= 3) >>> boolToList)
+reactimate $ concatS after3Empty  >>> arrM print
+"Yes"
+"Yes"
+"Yes"
+^CInterrupted.
 -}
 concatS :: Monad m => MStream m [b] -> MStream m b
 concatS msf = MSF $ \_ -> tick msf []
--- a/src/Data/MonadicStreamFunction/Async.hs
+++ b/src/Data/MonadicStreamFunction/Async.hs
@@ -1,0 +1,45 @@
+module Data.MonadicStreamFunction.Async where
+
+-- dunai
+import Data.MonadicStreamFunction.Core
+import Data.MonadicStreamFunction.Util (MStream)
+
+-- * Asynchronous operations on monadic stream functions
+
+{- |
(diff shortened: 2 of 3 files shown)


ivanperez-keera:
Also, why not make it polymorphic in the input?

 
```
-- | Concatenates a monadic stream of lists to a monadic stream.
concatS :: Monad m => MSF m a [b] -> MSF m a b
concatS msf = MSF $ tick msf []
  where
    tick msf (b:bs) a = return (b, MSF $ tick msf bs)
    tick msf []     a = do
      (bs, msf') <- unMSF msf a
      tick msf' bs a
```

turion:
> Just in case: have you thought about the alternative implementation in which the internal `MSF` steps at every sample, regardless of whether the accumulator is empty or not?

I think that doesn't work. Imagine an `MSF` that outputs a list of length 2 at every step. The accumulator would grow indefinitely.

ivanperez-keera:
Yes. Well, it works differently :) Your way, it introduces asynchronicity.

Would it be safe to say that, up until now, everything else is synchronous?

So far, don't all MSFs tick exactly once per sampling time, unless you access the constructor for MSFs.

turion:
> Also, why not make it polymorphic in the input?

That would be nice, but it looses data every time a nonempty list is produced.

If I had incoming data for every element of the list, I'd first apply mapMSF:
```
produceLists :: MSF m () [a]
processStuff :: MSF m a b
produceBs :: MSF m () b
produceBs = concatS $ produceLists >>> mapMSF processStuff
```

turion:
> Yes. Well, it works differently :) Your way, it introduces asynchronicity.
Would it be safe to say that, up until now, everything else is synchronous?
So far, don't all `MSF`s tick exactly once per sampling time, unless you access the constructor for `MSF`s.

Very good point. Yes, actually that's exactly the reason I implemented it. Using `concatS`, one can do more general things like asynchronously merging/zipping to `MSF`s etc.
Are you saying we should somehow document this difference?

ivanperez-keera:
>     Also, why not make it polymorphic in the input?
> 
> That would be nice, but it looses data every time a nonempty list is produced.

I don't get it. It's just a generalisation of your implementation, isn't it? You are just assuming that the input is of type `()`:

`yourConcatS = arr (const ()) >>> myConcatS `

turion:
> Also, both implementations have the precondition that the MSF must produce a non-empty list at every point

I don't think so. The only precondition is that the MSF must _eventually_ produce a non-empty list. If it starts producing empty lists _forever_, then yes, it hangs up.

ivanperez-keera:
> > Also, both implementations have the precondition that the MSF must produce a non-empty list at every point
> 
> I don't think so. The only precondition is that the MSF must eventually produce a non-empty list. If it starts producing empty lists forever, then yes, it hangs up.

There's a difference between forever/eventually at this conceptual sampling time, and forever/eventually at some future sampling time.

By _at every point_ I mean that it needs to produce some output without the monadic context changing due to other MSFs in the network.

turion:
> But this pull request includes the changes in #30 .

Oops, my bad. I'll have to pay more attention. Anyways, that's merged now.

turion:
> By _at every point_ I mean that it needs to produce some output without the monadic context changing due to other MSFs in the network.

I'm not sure I'm following entirely. Consider the following examples:
```
everyOtherEmpty = proc _ -> do
  n <- count -< ()
  returnA -< if even n
    then [ "Hello" ]
    else []

after3Empty = proc _ -> do
  n <- count -< ()
  returnA -< if n > 3
    then []
    else [ "Hello" ]
```
Do we agree that `concatS everyOtherEmpty` is ok and will be productive, while `concatS after3Empty` will hang up?

ivanperez-keera:
> Do we agree that `concatS everyOtherEmpty` is ok and will be productive, while `concatS after3Empty` will hang up?

I do.

> Using concatS, one can do more general things like asynchronously merging/zipping to MSFs etc.
> Are you saying we should somehow document this difference?

Definitely it should be documented.

And possibly even more. There two ways to classify this function: as a stream related function (although I think it can be generalised to MSFs), and as an asynchronous MSF function.

I'm wondering if all async stuff should be in an Async module. That's where I'd put it and where I'll put my long-overdue-to-be-pushed changes that also did asynchronous stuff.

turion:
> I'm wondering if all async stuff should be in an Async module. That's where I'd put it and where I'll put my long-overdue-to-be-pushed changes that also did asynchronous stuff.

Oh yes, very good. Pushed.

turion:
>> Do we agree that `concatS everyOtherEmpty` is ok and will be productive, while `concatS after3Empty` will hang up?

>I do.

I just checked, GHC does as well. And now the documentation even claims it! It must be true!

PR #35, opened 2017-10-18
Fix #33

### LINKED ISSUE #33: Lists to MSFs in MaybeT to exceptions
In languages that distinguish between inductive and coinductive types, lists are inductive types, while the corresponding coinductive type is a "Stream that can end", i.e. an `MStream Maybe`. Consequently, there should be a map `[a] -> MStream (MaybeT m) a` (or even `ListT m a -> MStream (MaybeT m) a` for a correct implementation of `ListT`).

`Maybe` is a special case of an exception, and there should be a utility `MSF (MaybeT m) a b -> MSF (ExceptT () m) a b`.

turion:
Fixes #33

### DIFF
--- a/src/Control/Monad/Trans/MSF/Except.hs
+++ b/src/Control/Monad/Trans/MSF/Except.hs
@@ -11,6 +11,7 @@ import qualified Control.Category as Category
 import Control.Monad.Trans.Class
 import Control.Monad.Trans.Except
   hiding (liftCallCC, liftListen, liftPass) -- Avoid conflicting exports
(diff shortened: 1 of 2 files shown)


ivanperez-keera:
Is there any reason why this should be merged into a release branch and not develop?

turion:
Depends on the branch model you want to follow. (I probably misunderstood it.) I was basing and PR'ing all issues with milestone 0.1.2 on this branch, but I can also PR to develop if that's the preferred workflow.

PR #37, opened 2017-10-21
Certain useful utilities that are sometimes needed when working with MSFExcept

turion:
When working on the `rhine` tutorial, it turned out that these are quite useful with exceptions.


ivanperez-keera:
We have used the name `step` for `stepMSF` (in the paper at least). Should `step` maybe be called `throwNext`?

turion:
`throwNext` is fine. I've also thought about calling it `tick`, but hesitated because of a name clash with Rhine.

PR #60, opened 2017-12-03
Contributing 'occasionally' and MonadRandom stuff to Dunai & BearRiver

thalerjonathan:
As discussed.

### DIFF
--- a/examples/bearriver/src/FRP/BearRiver.hs
+++ b/examples/bearriver/src/FRP/BearRiver.hs
@@ -156,16 +156,16 @@ after q x = feedback q go
               return ((e, t'), ct)
 
 occasionally :: MonadRandom m 
-            => Time -- ^ The time /q/ after which the event should be produced on average
-            -> b    -- ^ Value to produce at time of event
-            -> SF m a (Event b)
+             => Time -- ^ The time /q/ after which the event should be produced on average
+             -> b    -- ^ Value to produce at time of event
+             -> SF m a (Event b)
 occasionally tAvg b
   | tAvg <= 0 = error "dunai: Non-positive average interval in occasionally."
   | otherwise = proc _ -> do
-    r <- getRandomRS (0, 1) -< ()
-    dt <- timeDelta         -< ()
-    let p = 1 - exp (-(dt / tAvg))
-    returnA -< if r < p then Event b else NoEvent
+      r   <- getRandomRS (0, 1) -< ()
+      dt  <- timeDelta          -< ()
+      let p = 1 - exp (-(dt / tAvg))
+      returnA -< if r < p then Event b else NoEvent
  where
   timeDelta :: Monad m => SF m a DTime
   timeDelta = arrM_ ask
--- a/examples/bearriver/src/FRP/BearRiver.hs
+++ b/examples/bearriver/src/FRP/BearRiver.hs
@@ -16,8 +16,10 @@ import           Control.Arrow                as X
 import qualified Control.Category             as Category
 import           Control.Monad                (mapM)
 --import           Control.Monad.Reader
+import           Control.Monad.Random
 import           Control.Monad.Trans.Maybe
 import           Control.Monad.Trans.MSF
+import           Control.Monad.Trans.MSF.Random
 import           Data.Traversable             as T
 import           Data.Functor.Identity
 import           Data.Maybe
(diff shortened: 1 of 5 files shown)


src/Control/Monad/Trans/MSF/Random.hs
```diff
@@ -0,0 +1,50 @@
...
+  return ((g', b), runRandS msf' g')
+
+-- | Updates the generator every step but discharges the generator 
+evalRandS  :: (RandomGen g, Monad m) => MSF (RandT g m) a b -> g -> MSF m a b
+evalRandS msf g = runRandS msf g >>> arr snd
+
+getRandomS :: (MonadRandom m, Random b) => MSF m a b
+getRandomS = proc _ -> do
+  r <- arrM_ $ getRandom -< ()
+  returnA -< r
```
ivanperez-keera: Is this the same as `getRandomS = arrM_ getRandom` or `getRandomS = constant () >>> arrM_ getRandom`?

examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -153,6 +155,21 @@ after q x = feedback q go
                   ct = if t' < 0 then constant (NoEvent, t') else go
               return ((e, t'), ct)
 
+occasionally :: MonadRandom m 
+            => Time -- ^ The time /q/ after which the event should be produced on average
```
ivanperez-keera: Please align?

examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -153,6 +155,21 @@ after q x = feedback q go
...
+            -> SF m a (Event b)
+occasionally tAvg b
+  | tAvg <= 0 = error "dunai: Non-positive average interval in occasionally."
+  | otherwise = proc _ -> do
+    r <- getRandomRS (0, 1) -< ()
+    dt <- timeDelta         -< ()
+    let p = 1 - exp (-(dt / tAvg))
+    returnA -< if r < p then Event b else NoEvent
+ where
+  timeDelta :: Monad m => SF m a DTime
```
ivanperez-keera: Consistency in indentation?

examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -153,6 +155,21 @@ after q x = feedback q go
...
               return ((e, t'), ct)
 
+occasionally :: MonadRandom m 
+            => Time -- ^ The time /q/ after which the event should be produced on average
+            -> b    -- ^ Value to produce at time of event
+            -> SF m a (Event b)
+occasionally tAvg b
+  | tAvg <= 0 = error "dunai: Non-positive average interval in occasionally."
+  | otherwise = proc _ -> do
+    r <- getRandomRS (0, 1) -< ()
```
ivanperez-keera: Indent one level into? (add  2 spaces to the arrow block).

ivanperez-keera: And align the `<-`

src/Control/Monad/Trans/MSF/Random.hs
thalerjonathan: Yes, thanks for the hint. Fixed it.

examples/bearriver/src/FRP/BearRiver.hs
thalerjonathan: Done. So if splitting up types of a function into multiple lines, they all should align under the :: ?

examples/bearriver/src/FRP/BearRiver.hs
thalerjonathan: Do you mean the where? I looked at the other code in BearRiver.hs and found that where was indented with 1 space. timeDelta is not as far indented as the above code because of guards.

examples/bearriver/src/FRP/BearRiver.hs
thalerjonathan: Not sure what you mean with one level into 2 spaces to the arrow block: it is already on level 2 (has 4 spaces). Added another level (6 spaces now).

examples/bearriver/src/FRP/BearRiver.hs
ivanperez-keera: You're right about the `where`. This is my fault. It's one of those places in which I tried to apply an indentation rule and, further down the line, I realised it was no good.

I tend to use two spaces:
```haskell
f x y z =
  someExpression x z y
```
That also applies to `where` clauses:

```haskell
f x y z = someExpression z'
  where
    z' = anotherExpression x y z 
```
The problem appears because I tend to put `where` to the left of the expression that uses it. That means I need to add 4 spaces on top, or only one for `where`:
```
f x y z =
    someExpression z'
  where
    z' = anotherExpression x y z 
```
Or
```
f x y z =
 someExpression z'
where
 z' = anotherExpression x y z 
```
All of which look wrong to my eye :( The second has the added issue that lhs2TeX won't accept it as properly indented haskell, and everything will be aligned to the left as top-level expressions.

So, I'll accept it as is and later do my job at making the syntax more consistent across the whole project, which I haven't done properly so far :|

Thanks!

Issue #40, opened 2017-10-22
`bearriver`: definitions missing


ivanperez-keera:
BearRiver is missing some definitions available in Yampa. Ideally, and this is a big issue, it should include everything.

One should not worry at all about optimisations. Leaks and similar issues should be monitored and addressed systematically later.

thalerjonathan:
I am currently experimenting with refactoring my code-base to Bearriver and need primitives like 'after', 'occationally' and 'repeatedly'. I will implement them and can then add them into Bearriver (and/or Dunai) if required.

turion:
I can recommend using exceptions for that. E.g. `repeatedly` should go something like this:
```
import Control.Monad.Trans.MSF.Except
repeatedly diff b = safely $ do
  _ <- try $ proc _ -> do 
     time <- integrate    -< 1
     _      <- throwOn () -< time > diff
     step $ return b
     safe $ repeatedly diff b
```

thalerjonathan:
I just realised that 'after' is implemented already.

Regarding 'occasionally': Yampa gets its stream of random-numbers from a RandomGen which is passed as an argument to occasionally. With MSFs it would be much more interesting to have a RandT monad somwhere in the stack from which to draw the random-numbers. I guess fixing it in the types would be too restrictive / wouldnt work as we fix the monad stack. A solution would be to have a function which generates a random-number from the existing monad stack - is this possible?

turion:
A clean solution to this would maybe be the following API:

```haskell
module Control.Monad.Trans.MSF.Rand

-- MonadRandom
import Control.Monad.Random

-- | Updates the generator every step
runRandS :: MSF (RandT g m) a b -> g -> MSF m a (g, b)

evalRandS  :: MSF (RandT g m) a b -> g -> MSF m a b
evalRandS msf g = runRandS msf g >>> arr snd
```

turion:
Getting a stream of random numbers is then as easy as calling `arrM_ getRandom`. There should probably be an alias for it then, called `getRandomS :: (Random b, MonadRandom m) => MSF m a b`.

thalerjonathan:
Here is my naive implementation of occasionally in Yampa style (passing in RandomGen explicitly): 

`occasionally :: (RandomGen g, Monad m) => g -> Time -> b -> SF m a (Event b)
occasionally g t_avg b
    | t_avg > 0 = MSF (const $ tf g b)
    | otherwise = error "AFRP: occasionally: Non-positive average interval."
[...]

PR #63, opened 2017-12-04
Added additional definitions from Yampa to BearRiver as discussed in …

thalerjonathan:
…in #40: never, now, once and >--

### DIFF
--- a/examples/bearriver/src/FRP/BearRiver.hs
+++ b/examples/bearriver/src/FRP/BearRiver.hs
@@ -143,6 +143,24 @@ hold a = feedback a $ arr $ \(e,a') ->
 loopPre :: Monad m => c -> SF m (a, c) (b, c) -> SF m a b
 loopPre = feedback
 
+-- | Event source that never occurs.
+never :: Monad m => SF m a (Event b)
+never = MSF $ \_ -> return (NoEvent, never)
+
+-- | Event source with a single occurrence at time 0. The value of the event
+-- is given by the function argument.
+now :: Monad m => b -> SF m a (Event b)
+now b0 = Event b0 --> never
+
+-- | Suppress all but the first event.
+once :: Monad m => SF m (Event a) (Event a)
+once = takeEvents 1
+
+-- | Suppress all but the first n events.
+takeEvents :: Monad m => Int -> SF m (Event a) (Event a)
+takeEvents n | n <= 0 = never
+takeEvents n = dSwitch (arr dup) (const (NoEvent >-- takeEvents (n - 1)))
+
 after :: Monad m
       => Time -- ^ The time /q/ after which the event should be produced
       -> b    -- ^ Value to produce at that time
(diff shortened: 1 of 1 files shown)


examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -143,6 +143,24 @@ hold a = feedback a $ arr $ \(e,a') ->
 loopPre :: Monad m => c -> SF m (a, c) (b, c) -> SF m a b
 loopPre = feedback
 
+-- | Event source that never occurs.
+never :: Monad m => SF m a (Event b)
+never = MSF $ \_ -> return (NoEvent, never)
```
ivanperez-keera:
```haskell
never = constant NoEvent
```

In general it's best not to use internal constructors if possible.

Using the API, as opposed to the internal constructors, facilitates changing the library internally without affecting uses/users. This is part of the reason for an API in the first place.

While this often creates some (small) performance loss, it's worth it when libraries grow big, since you can change the internals quickly. For example, if you choose to use GADTs to improve performance, and implement `constant` using that optimisation, you'll only get a gain if you stuck to the API. For another example, if you choose to add another field or change the order of the elements in the tuple, you won't have to change everything.

examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -170,11 +188,26 @@ occasionally tAvg b
...
 
+-- | Input initialization operator.
+--
+-- The input at time zero is the first argument, and from
+-- that point on it behaves like the signal function passed as
+-- second argument.
+(>--) :: Monad m => a -> SF m a b -> SF m a b
+a0 >-- sf = MSF $ \_ -> do
+  (b, ct) <- unMSF sf a0
+  return (b, ct)
```
ivanperez-keera: This could be factorized:
```haskell
replaceOnce :: a -> MSF a a -> MSF a a
replaceOnce x sf = MSF $ \_ -> unMSF x

(-->) :: Monad m => b -> SF m a b -> SF m a b 
b0 --> sf = sf >>> replaceOnce b0

(>--) :: Monad m => a -> SF m a b -> SF m a b 
a0 >-- sf = replaceOnce a0 >>> sf
```

The question is: can one write `replaceOnce` without using `unMSF`? It's possible, using switching :)

ivanperez-keera: @turion You've implemented a new exception-handling mechanism. What's the current way of writing `\x -> dSwitch (constant (x, Event)) (\_ -> id)` using that API?

examples/bearriver/src/FRP/BearRiver.hs
thalerjonathan: Yes, I understand.
Saw that you use the internal MSF constructor a couple of times and assumed it is OK in such cases. But of course when a function can be implemented using the existing API then it should always go that way - but sometimes it is not clear for me to see that but thanks for the help.

examples/bearriver/src/FRP/BearRiver.hs
thalerjonathan: Awesome :)
Shouldn't ```replaceOnce``` look like: 
```haskell
replaceOnce :: Monad m => a -> MSF m a a
replaceOnce x = MSF $ \_ -> return (x, arr id)
```

thalerjonathan: And here without using the internal MSF constructor:
```haskell
replaceOnce :: Monad m => a -> SF m a a
replaceOnce a = dSwitch (arr $ const (a, Event ())) (const $ arr id)
```

ivanperez-keera:
```haskell
replaceOnce :: Monad m => a -> SF m a a
replaceOnce a = dSwitch (constant (a, Event ())) (const $ arr id)
```

?

thalerjonathan: Yeah obviously ;)

PR #64, opened 2017-12-04
Implemented a new feature: uniform super-sampling for BearRiver. Uses…

thalerjonathan:
… internal MSF constructor for now

### DIFF
--- a/examples/bearriver/src/FRP/BearRiver.hs
+++ b/examples/bearriver/src/FRP/BearRiver.hs
@@ -188,6 +188,43 @@ occasionally tAvg b
   timeDelta :: Monad m => SF m a DTime
   timeDelta = arrM_ ask
 
+-- | Performs uniform super sampling on the provided signal-function with n samples
+-- 
+-- This is done by dividing the current time-delta into N equally spaced samples 
+-- and evalutating the signal function with these subsamples and always the same input. 
+-- The first argument is the number of samples - if LT 0 exactly 1 sample is calculated
+-- The second argument is the signal-function to sample
+-- Returns a new signal-function which returns a list of the super-samples
+superSamplingUniform :: Monad m => Int -> SF m a b -> SF m a [b]
+superSamplingUniform n sf = MSF $ \a -> do
+  dt        <- ask
+  (sf', bs) <- superSampleRun n dt sf a
+  return (reverse bs, superSamplingUniform n sf') -- reverse because of accumulator
+ where
+  superSampleRun :: Monad m 
+                 => Int 
+                 -> DTime 
+                 -> SF m a b 
+                 -> a 
+                 -> ReaderT DTime m (SF m a b, [b])
+  superSampleRun n dt sf a 
+      | n <= 1 = superSampleMulti 1 dt sf a []
+      | otherwise = superSampleMulti n superDt sf a []  
+    where
+      superDt = dt / fromIntegral n
+
+  superSampleMulti :: Monad m 
+                   => Int 
+                   -> DTime 
+                   -> SF m a b 
+                   -> a 
+                   -> [b]
+                   -> ReaderT DTime m (SF m a b, [b])
+  superSampleMulti 0 _ sf _ acc = return (sf, acc)
+  superSampleMulti n superDt sf a acc = do
+    (b, sf') <- unMSF sf a
+    superSampleMulti (n-1) superDt sf' a (b:acc) 
+
 -- | Initialization operator (cf. Lustre/Lucid Synchrone).
 --
 -- The output at time zero is the first argument, and from


examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -188,6 +188,43 @@ occasionally tAvg b
   timeDelta :: Monad m => SF m a DTime
   timeDelta = arrM_ ask
 
+-- | Performs uniform super sampling on the provided signal-function with n samples
+-- 
+-- This is done by dividing the current time-delta into N equally spaced samples 
+-- and evaluating the signal function with these subsamples and always the same input. 
+-- The first argument is the number of samples - if LT 0 exactly 1 sample is calculated
```
turion: Can you rather use haddock function argument comments? See e.g. https://www.haskell.org/haddock/doc/html/ch03s02.html#idm140354810892448 (3.2.3. Function arguments)

turion:
Cool idea! You should be able to get rid of the `MSF` constructor by using `Data.MonadicStreamFunction.Util.mapMSF`. It's one of the fundamental asynchronous operators on `MSF`s (the other one being `Data.MonadicStreamFunction.Async.concatS`), and can be used for such purposes. I'm guessing an implementation like this:

```haskell
superSamplingUniform n msf
  | n <= 1    = msf
  | otherwise = readerS $ proc (dTime, a) -> mapMSF (runReaderS msf) -< [ (i * dTime / n, a) | i <- [1..n] ]
```

thalerjonathan:
Thanks for the hint, its an amazingly elegant solution! It took me a while to understand and get it right:

```haskell
superSamplingUniform :: Monad m 
                     => Int          -- ^ Number of samples into which the current time-delta is being uniformly subdivided. If LT 0, exactly 1 sample will be generated
                     -> SF m a b     -- ^ The signal-function to sample
                     -> SF m a [b]   -- ^ The new signal-function which performs the super sampling, length of [b] is number of samples (except in case of LT 0, then its 1).
superSamplingUniform n sf 
  | n <= 1    = proc a -> do
    b <- sf -< a
    returnA -< [b]
  | otherwise = readerS (proc (dt, a) -> do
    let superDt = dt / fromIntegral n
    let dtas = replicate n (superDt, a)
    mapMSF (runReaderS sf) -< dtas)
```

turion:
Ah yes, you're right! You can slightly simplify the `n <= 1` case if you want: `sf >>> arr return`. (Since `return :: a -> [a]` is just `\a -> [a]`.)

examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -191,47 +191,19 @@ occasionally tAvg b
...
-                   -> ReaderT DTime m (SF m a b, [b])
-  superSampleMulti 0 _ sf _ acc = return (sf, acc)
-  superSampleMulti n superDt sf a acc = do
-    (b, sf') <- unMSF sf a
-    superSampleMulti (n-1) superDt sf' a (b:acc) 
+superSamplingUniform n sf 
+  | n <= 1    = proc a -> do
+    b <- sf -< a
+    returnA -< [b]
+  | otherwise = readerS (proc (dt, a) -> do
```
turion: How about `readerS $ ...`?

examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -205,7 +220,7 @@ b0 --> sf = sf >>> replaceOnce b0
 a0 >-- sf = replaceOnce a0 >>> sf
 
 replaceOnce :: Monad m => a -> SF m a a
-replaceOnce a = dSwitch (arr $ const (a, Event ())) (const $ arr id)
+replaceOnce a = dSwitch (constant (a, Event ())) (const $ arr id)
```
turion: You can refactor `const $ arr id` as well, I think. It should even work with `Category.id`.

examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -188,6 +188,21 @@ occasionally tAvg b
   timeDelta :: Monad m => SF m a DTime
   timeDelta = arrM_ ask
 
+-- | Performs uniform super sampling on the provided signal-function with n samples
+-- 
+-- This is done by dividing the current time-delta into N equally spaced samples 
+-- and evalutating the signal function with these subsamples and always the same input. 
+superSamplingUniform :: Monad m 
+                     => Int          -- ^ Number of samples into which the current time-delta is being uniformly subdivided. If LT 0, exactly 1 sample will be generated
```
turion: Better break too long lines.

examples/bearriver/src/FRP/BearRiver.hs
turion: Just a note: It seems like this change is unrelated to your pull request.

ivanperez-keera:
There are several ways of doing re-sampling (I'm not arguing for or against this, just trying to have a discussion about what to introduce).

For example: https://github.com/ivanperez-keera/dunai/blob/develop/examples/timetransform/TimeTransform.hs#L103

examples/bearriver/src/FRP/BearRiver.hs
ivanperez-keera: Good point. And, yes, it helps to keep PR straight to the point and focus on one feature only. It makes the maintainer live's so much easier.

thalerjonathan: Yes you got me there, I thought I could sneak it in as general "refactoring". Will do it next time properly, sorry.

thalerjonathan:
Ok then what about expressing uniform super-sampling in terms of your resample function?
Also I would like implement a random-noise super sampling (inspired from computer graphics where it is used to soften shadow-edges).

ivanperez-keera:
This is becoming big-ish. I suggest the following:

1) Add it to a branch or to `examples/timetransform`, merge our efforts, and try, together, to come up with a general, clean solution that combines what you want and what was in `examples/timetransform`. If branch, I suggest the name `develop_resampling`.

2) Optionally, create a separate PR later for  `random-noise super sampling`, unless that is intuitively built by combining your last two PRs.

turion:
Hmmm... you know I've written a whole package based on dunai just about resampling and scheduling, right? (I'm talking about [this one](https://github.com/turion/rhine)).

ivanperez-keera:
Yes.

Rhine requires type-level clock coordination, which is not always what you want. However, the implementation might even be quite similar, except its a _proof_ of a different type (formula). There's also the possibility that there will be less need for auxiliary types/functions because of that.

Time transformations and re-sampling exist in Yampa, which has no notion of type-level clocks. For now, these have been added to a branch.

thalerjonathan:
@ivanperez-keera I will stick to my simple implementation for now as I need to move forward with making the transition to Dunai / BearRiver but will come back at a later point.

@turion your rhine library looks very interesting, it could be used for scheduling of my agents on a type-level. Also I realised that my approach to agent-based simulation also allows to implement System Dynamics, so I am now looking for a way to add Discrete Event Simulation (DES) which is the 3rd of the simulation methodologies - maybe rhine could be of help there.
Does rhine allow a random scheduling or is it always fixed at compile-time on the type-level? 
In the examples I found the notion of millisecond clocks, are there also logical clocks with continuous time?

PR #67, opened 2017-12-04
Fix #66

### LINKED ISSUE #66: Add NormedSpace instances (e.g. for tuples)
Either the issue in [Data/VectorSpace.hs, line 65](https://github.com/ivanperez-keera/dunai/blob/472d5d23ab758ff4d8ac9a55885895d6c622f125/src/Data/VectorSpace.hs#L65) should be fixed (basically an `instance (Floating (Groundfield v), VectorSpace v, InnerProductSpace v) => NormedSpace v`), or we should implement `NormedSpace` instances for tuples in `Data.VectorSpace.Tuples`.

turion:
Note that this does a little more than just fix #66. But I think the rest is long necessary cleanup.

Question: Is it necessary to haddock-comment instances in code?

### DIFF
--- a/src/Data/VectorSpace.hs
+++ b/src/Data/VectorSpace.hs
@@ -1,6 +1,6 @@
-{-# LANGUAGE TypeFamilies           #-}
-{-# LANGUAGE FlexibleContexts       #-}
-{-# LANGUAGE UndecidableInstances   #-}
+{-# LANGUAGE TypeFamilies      #-}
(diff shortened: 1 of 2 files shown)


ivanperez-keera:
I could be wrong, but I think haddock comments for instances translate into nothing.

In one of our projects we listed the instances in a module to get at least something: it was a module with just instances, so otherwise the haddock documentation would have been empty.

PR #68, opened 2017-12-05
add `mealy`

ggreif:
See https://en.wikipedia.org/wiki/Mealy_machine

### DIFF
--- a/src/Data/MonadicStreamFunction/Util.hs
+++ b/src/Data/MonadicStreamFunction/Util.hs
@@ -136,6 +136,14 @@ accumulateWith f s0 = feedback s0 $ arr g
   where
     g (a, s) = let s' = f a s in (s', s')
 
+-- | Applies a transfer function to the input and an accumulator,
+-- returning the updated accumulator and output. Equal to
+-- @\f s0 -> feedback s0 $ arr (uncurry f)@.
+mealy :: Monad m => (a -> s -> (b, s)) -> s -> MSF m a b
+mealy f s0 = feedback s0 $ arr g
+  where
+    g (a, s) = let (b, s') = f a s in (b, s')
+
 -- * Unfolding
 
 -- | Generate outputs using a step-wise generation function and an initial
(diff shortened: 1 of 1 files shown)


src/Data/MonadicStreamFunction/Util.hs
```diff
@@ -136,6 +136,14 @@ accumulateWith f s0 = feedback s0 $ arr g
   where
     g (a, s) = let s' = f a s in (s', s')
 
+-- | Applies a transfer function to the input and an accumulator,
+-- returning the updated accumulator and output. Equal to
+-- @\f s0 -> feedback s0 $ arr (uncurry f)@.
+mealy :: Monad m => (a -> s -> (b, s)) -> s -> MSF m a b
+mealy f s0 = feedback s0 $ arr g
+  where
+    g (a, s) = let (b, s') = f a s in (b, s')
```
turion: Is there any reason not to write `arr $ uncurry f` directly?

ggreif: @turion One would have to examine Core/STG output to give a good judgment.

ivanperez-keera: I suggest opting for the simplest, most declarative solution possible by default, and only optimising when there is a real gain.

turion: I see, so it might be possible that when writing `arr $ uncurry g$, it will expand `$` and `uncurry` on every tick? Seems plausible...

But what's the benefit of the `let`-construction?

turion: @ggreif , Ivan will probably start a research project on measuring and optimising performance of FRP, so maybe we'll gain some insights then.

ivanperez-keera: @turion I'm in a similar position: started the research project and I'm evaluating the performance of different FRP constructs.

ggreif:
A `moore` machine would be interesting to be added too, but currently I have no use for it in my FRP projects. In [`CLaSH`](https://github.com/clash-lang/clash-compiler) OTOH, I use them pretty often.

Issue #61, opened 2017-12-04
Make travis upload new versions of both dunai and bearriver to hackage


ivanperez-keera:
In order to simplify our lives, we could make travis automatically publish stable versions of packages on hackage. A new user has been created for this ( #59 ).

Example of how to do this is are in https://github.com/ivanperez-keera/Yampa/blob/develop/.travis.yml#L31-L38, and https://github.com/keera-studios/keera-hails/blob/master/.travis.yml#L86-L109.

turion:
Stupid question. Is it not a problem that the password is public then?

Btw, if you want to avoid duplicating the list of subpackages every time, you can do the thing described in https://docs.travis-ci.com/user/languages/haskell/#Build-Matrix.

ivanperez-keera:
It won't be public. It'll be encrypted by Travis. But it'll be visible by anyone with commit access to the repo, which is why it's best to have a different user for this.

ivanperez-keera:
I just took a quick look. It seems to me that, since packages need to be compiled with a range of GHC versions, the cleanest way to do this would be to specify a build matrix that provides all possible combinations:
```
env:
 - GHCVER=7.6.3  PACKAGE="dunai"     PACKAGE_DIR="."
 - GHCVER=7.8.4  PACKAGE="dunai"     PACKAGE_DIR="."
 - GHCVER=7.10.3 PACKAGE="dunai"     PACKAGE_DIR="."
 - GHCVER=8.0.2  PACKAGE="dunai"     PACKAGE_DIR="."
 - GHCVER=8.2.1  PACKAGE="dunai"     PACKAGE_DIR="."

 - GHCVER=7.6.3  PACKAGE="bearriver" PACKAGE_DIR="examples/bearriver"
 - GHCVER=7.8.4  PACKAGE="bearriver" PACKAGE_DIR="examples/bearriver"
 - GHCVER=7.10.3 PACKAGE="bearriver" PACKAGE_DIR="examples/bearriver"
 - GHCVER=8.0.2  PACKAGE="bearriver" PACKAGE_DIR="examples/bearriver"
 - GHCVER=8.2.1  PACKAGE="bearriver" PACKAGE_DIR="examples/bearriver"
```
Is there a simpler way? (yes, we can eliminate `PACKAGE_DIR` if we put dunai in a subdir and bearriver in another, but that's not what I am asking.)

chriz-keera:
I'm not sure about this, but we could try the following or a variant of it:

```
[...]

PR #73, opened 2017-12-07
Adds initial setup for dunai deploy to hackage.

chriz-keera:
Adds username, encrypted password and conditions for deploy.

Refs #61.


ivanperez-keera:
Thank you, @chriz-keera  :)

Issue #74, opened 2017-12-07
Bearriver does not compile with GHC 7.6


ivanperez-keera:
I'm not sure we still want to support GHC 7.6, but bearriver does not currently compile with GHC 7.6 simply because the constraint on base is `>=4.7`, not `>=4.6`.

Just to be consistent with the current `.travis.yml` file, we could support GHC 7.6.

This merely requires changing the lower boundary in bearriver's cabal file.

ivanperez-keera:
I suggest we keep it around for as long as it lasts, but don't make much of an effort to be backwards compatible with 7.6. I would prefer to support 7.8 for now.

I'm not sure how to make a sensible decision about how far back to go. The lowest I'm using for Android right now is 7.10, although I have some 7.8 apps lying around.

turion:
Sounds good to me. Just support a particular version until it would become an effort to support it, at which point it can be dropped. We could add a line in the README or elsewhere to mention the oldest GHC version that is still supported.

ivanperez-keera:
Sounds good to me to have that line upfront so that users know what to expect.

ivanperez-keera:
Actually, until recently I was using a version of GHC without the Functor-Monad hierarchy and dunai wouldn't compile for me unless I changed that by hand.

Issue #58, opened 2017-11-30
Test examples with travis


turion:
We should build all the examples with travis as well. As inspiration for a `.travis.yml` that works for several cabal packages in the same repository, have a look [here](https://github.com/turion/rhine/blob/dev_rhine_gloss/.travis.yml).

ivanperez-keera:
You may have changed things. Do you mean the current? https://github.com/turion/rhine/blob/master/.travis.yml

Thanks for providing that file. This is useful. Is this possibility of compiling multiple packages by changing PACKAGE in the environment in the build matrix documented anywhere in Travis' documentation?

And a question I'm pondering: should bearriver compile against the version of dunai in the repo or the one on hackage? And what does it mean in each case? And how to do that with the generality you use in that travis file? (`for i in $(find ../ -iname '*.cabal' | xargs -n 1 dirname); do cabal sandbox add-source $i; done` ?)

turion:
This feature is documented since a few days (I added it :grin:) in https://docs.travis-ci.com/user/languages/haskell/#Build-Matrix. It might mean that `dunai` needs to go to a subdir as well.

I think `bearriver` should compile against the hackage version, because the typical use case is that people will create a new project and just put `bearriver` and `dunai` in the dependencies, which gets them from hackage. Coincidentally, this also simplifies the build process, since we don't need to use `cabal sandbox add-source`.

ivanperez-keera:
Agreed.

turion:
So can we add the other example packages to the travis build as well now?

ivanperez-keera:
Yes.

We can add those that we are ready to make a compromise to maintain.

The travis file should not push those to hackage, though. This needs to be adjusted in the deploy condition.

turion:
IMHO, we should maintain all examples, or remove them from the repo.

turion:
Any opinion about how to proceed with version numbers in the examples? Two possibilities come to my mind:

* Treat all example packages individually, i.e. let each one have its own version numbers
* Bump version numbers synchronously with dunai (I do that with rhine and its subsidiary packages)

ivanperez-keera:
> Any opinion about how to proceed with version numbers in the examples? Two possibilities come to my mind:
> 
>     Treat all example packages individually, i.e. let each one have its own version numbers
>
>     Bump version numbers synchronously with dunai (I do that with rhine and its subsidiary packages)

[...]

PR #75, opened 2017-12-07
Publishes stable versions of dunai and bearriver on hackage.

chriz-keera:
Travis should automatically publish stable versions of dunai and bearriver packages on hackage (Refs #61).

This commit modifies the travis config to:

  - Use the standard building image for Haskell.
  - Compile both dunai and bearriver.
  - Deploy both libraries to hackage (only for latest version of GHC).

The boundary on base for bearriver was lowered for compatibility with GHC 7.6 (Refs. #74).


turion:
When exactly will it deploy? It would be wrong to deploy on every push to master that builds. Does it somehow know when the version numbers go up?

If you want to automatically deploy, then bearriver should maybe build with the git dunai (i.e. `cabal sandbox add-source ../..` or so) and not hackage dunai. Otherwise you'd have to bump dunai and bearriver version numbers after each other, or bearriver will never build since it can't find the updated dunai on hackage yet.

ivanperez-keera:
This is exactly why we have the master-develop policy. Master means ready to release. If it's not ready to release, we should not commit to master.

Hackage will prevent the same version from being uploaded twice. Unfortunately, we have no (easy) way of knowing that ahead, although `cabal list` might give us a hint. The last I tried seemed a bit hacky. Improvements in this respect are very welcome. (I'm having a similar problem deploying to ubuntu repos.)

You're right about bearriver compiling against the version on hackage, not the repo. This is why in #58 I asked:

> And a question I'm pondering: should bearriver compile against the version of dunai in the repo or the one on hackage? And what does it mean in each case? And how to do that with the generality you use in that travis file? (`for i in $(find ../ -iname '*.cabal' | xargs -n 1 dirname); do cabal sandbox add-source $i; done` ?)

To which you responded:

> I think bearriver should compile against the hackage version, because the typical use case is that people will create a new project and just put bearriver and dunai in the dependencies, which gets them from hackage.

I have not found an ideal solution yet. One provides consistency for hackage users, and another consistency for the repo.

I'm happy closing this for now, knowing that it is an improvement over what we had. Worst case scenario would be that restarting the travis jobs we'd get to consistency both for the repo and hackage.

This problem applies generally to any project with more than one library in the same repo, so I'm not going to open an issue about it unless one of us is going to work on it right away.

Things that might help in the future:
  - Knowing if a version has been uploaded.
  - Establishing order between combinations in the travis build matrix (we could delay bearriver until we know dunai has successfully uploaded.)
  - Uploading bearriver only when dunai compiles and uploads first (but that's incomplete in a different way).
  - Eventually, moving bearriver to its own repo (but that's like restarting all the travis job when bearriver fails to upload).

turion:
> This is exactly why we have the master-develop policy. Master means ready to release. If it's not ready to release, we should not commit to master.

Makes sense. (And hotfixes should increment the smallest version number.)

> Hackage will prevent the same version from being uploaded twice. Unfortunately, we have no (easy) way of knowing that ahead, although cabal list might give us a hint.
Ah, and will the travis build fail in that case?


> [bearriver depending on hackage dunai vs. repo dunai]
> I have not found an ideal solution yet. One provides consistency for hackage users, and another consistency for the repo.
If the automatic deploy works fine, one is as good as the other.

> Uploading bearriver only when dunai compiles and uploads first (but that's incomplete in a different way).
What's the problem there? First push the dunai version bump, wait until it deployed, then push all dependent repos. It's a bit inconvenient, but should work in principle, right?

PR #82, opened 2017-12-08
Fixes documentation in a lot of places. Fixes #51.

### LINKED ISSUE #51: Cross-reference and code highlight documentation
We should add cross-references in the haddock, e.g. writing `'MSF'` in most places instead of `MSF` or `@MSF@`. Where no useful reference exist, still prefer monospace `@someHaskellConcept@` over someHaskellConcept.

turion:


src/Control/Monad/Trans/MSF/Reader.hs
```diff
@@ -1,14 +1,14 @@
...
 
--- | MSFs with a Reader monadic layer.
+-- | 'MSF's with a 'Reader' monadic layer.
 --
--- This module contains functions to work with MSFs that include a 'Reader'
--- monadic layer. This includes functions to create new MSFs that include an
--- additional layer, and functions to flatten that layer out of the MSF's
+-- This module contains functions to work with 'MSF's that include a 'Reader'
+-- monadic layer. This includes functions to create new 'MSF's that include an
+-- additional layer, and functions to flatten that layer out of the 'MSF''s
```
ivanperez-keera: Not a big deal, but this will render incorrectly (the double single quote).

turion:
Do you know how to format the ' correctly?

ivanperez-keera:
According to https://github.com/haskell/haddock/issues/162, you need to write it like this: ```'MSF`'s```.

This happens in three modules (`Reader`, `State` and `Writer`).

turion:
Thanks for the hint!

turion:
Merging since the remaining minor formatting issue is taken care of.

Issue #26, opened 2017-09-17
Reorg: Function naming


ivanperez-keera:
The nomenclature used for the API is not intuitive. Why are is there a function called `arrM`, another called `arr` and another `liftS`, when they all have similar signatures?

This needs to be improved.

The naming convention needs to be such that the entry barrier to entry is minimal. I'd rather use names related to signal processing than operational notions (step), but the latter may be unavoidable.

turion:
I'm proposing the following naming guidelines:

* Suffix `M` generalises something to monads. Example: `arrM :: (a -> m b) -> MSF m a b`
* Suffix `S` generalises something to streams or stream functions. Example: `concatS :: MStream m [b] -> MStream m b`
* Suffix `_` specialises to something not having an input. Example: `arrM_ :: m b -> MSF m a b`
* `lift` means lifting into a monad transformer
* `liftBase` means lifting into a monad base
* `hoist` means applying a (general) monad morphism
* `run` means removing a monad transformer layer

This way, all the names follow a clear and standard pattern.
It would imply the following (and maybe other) renames:

* `liftS` -> `arrBase :: (Monad m2, MonadBase m1 m2) => (a -> m1 b) -> MSF m2 a b`
* `liftMSFPurer` -> `hoistS :: (Monad m2, Monad m1) => (forall c . m1 c -> m2 c) -> MSF m1 a b -> MSF m2 a b`
* `liftMSFTrans` -> `liftS :: (MonadTrans t, Monad m, Monad (t m)) => MSF m a b -> MSF (t m) a b`
* `liftMSFBase` -> `liftBaseS :: (Monad m2, MonadBase m1 m2) => MSF m1 a b -> MSF m2 a b`

This would break some code in articles and @ivanperez-keera 's thesis, though.

If you're fine with this, I'll submit a pull request.

ivanperez-keera:
A few comments:

* It's ok if my thesis breaks. Ideally it shouldn't at the time of submission, but eventually it will. We can't hold back progress because I wrote a thing.

   I'd rather just update the examples in the final submission if the reviewers agree, or add a file to some repo with the examples in the thesis for the updated API.

* My intuition is that if something ends in `_`, the type is `h ()` for some `h` (probably an `MSF m a ()`). For instance, I'd call `arrM_` `constM`.

[...]

PR #84, opened 2017-12-09
Generalised lifting. Fixes #16.

### LINKED ISSUE #16: Reorg: unify transformer lifting and running
@ivanperez-keera, we have a lot of different lifting and running functions around, do you have an opinion on what we should keep and what should be removed? I think we should go with one of the general approaches and reformulate (with inlining) everything in terms of it.

turion:
Rewrites most lifting/hoisting functions in terms of a single one. The ones that can't be rewritten in this approach are those that influence control flow.

Also removes a lot of alternative lifting functions and corresponding discussions. Come back to this point in history in case you want to compare performance of different approaches.


turion:
This fixes most of #16. Handling control flow requires more research and should be a separate issue, to be opened when that research is done.

ivanperez-keera:
This merge includes a function `hoistGen` and is pending the name discussion in https://github.com/ivanperez-keera/dunai/issues/26

ivanperez-keera:
This branch will be kept, but the pull denied to continue working on a slightly different issue it presents.

Merging into develop now partially depends on: https://github.com/ivanperez-keera/dunai/issues/26.

PR #88, opened 2017-12-29
following issue #40 renamed existing implementation of occasionally w…

thalerjonathan:
…hich requires MonadRandom type-class to occasionallyM and implemented a new occasionally which is an exact re-implementation of Yampas occasionally which draw its random-numbers from a RandomGen provided through an argument. Also implemented noise and noiseR as in Yampa which is used by the occasionally implementation.

### DIFF
--- a/examples/bearriver/src/FRP/BearRiver.hs
+++ b/examples/bearriver/src/FRP/BearRiver.hs
@@ -176,20 +176,54 @@ after q x = feedback q go
                   ct = if t' < 0 then constant (NoEvent, t') else go
               return ((e, t'), ct)
 
-occasionally :: MonadRandom m 
-             => Time -- ^ The time /q/ after which the event should be produced on average
-             -> b    -- ^ Value to produce at time of event
+occasionally :: (Monad m, RandomGen g) 
+             => g 
+             -> Time 
+             -> b 
              -> SF m a (Event b)
-occasionally tAvg b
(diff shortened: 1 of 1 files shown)


turion:
Somehow, the first half of your comment got deleted?

examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -176,20 +176,54 @@ after q x = feedback q go
...
-occasionally :: MonadRandom m 
-             => Time -- ^ The time /q/ after which the event should be produced on average
-             -> b    -- ^ Value to produce at time of event
+occasionally :: (Monad m, RandomGen g) 
+             => g 
+             -> Time 
+             -> b 
              -> SF m a (Event b)
-occasionally tAvg b
+occasionally g tAvg b
```
turion: I think you should be able to write something like:
```haskell
occasionally g tAvg b = runRandS (occasionally tAvg b) g >>> arr fst
```
If this works, I suggest rewriting the other primitives in the same way.

thalerjonathan:
Ah sorry, didn't realise that commits to my fork are automatically part of the pull request as well. I am currently out of time to commit more work to these changes but hope to find some time in the next weeks.

ivanperez-keera:
EDIT: Changed title, merged old title in first comment.

PR #89, opened 2017-12-29
Make ListT usable and fix #36

### LINKED ISSUE #36: Use a better ListT
The `ListT` implementation we're using (from `transformers`) is wrong (it's just `m [a]`) and we should eventually settle for a different, correct version.

turion:
This revamps the `Control.Monad.Trans.MSF.List` module and uses the more popular `ListT` monad.

* New (light) dependency `list-t`
* Improves naming and documentation in `Control.Monad.Trans.MSF.List`
* Adds module to package

### DIFF
--- a/dunai.cabal
+++ b/dunai.cabal
@@ -57,6 +57,7 @@ library
   exposed-modules:   Control.Monad.Trans.MSF
                      Control.Monad.Trans.MSF.Except
                      Control.Monad.Trans.MSF.GenLift
+                     Control.Monad.Trans.MSF.List
                      Control.Monad.Trans.MSF.Maybe
                      Control.Monad.Trans.MSF.Random
                      Control.Monad.Trans.MSF.Reader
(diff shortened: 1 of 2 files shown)


turion:
Whoops, I'm just noticing that this code still uses the `MSF` constructor. Feel free not to merge if this is a bother. (I'm not sure how to fix this, though)

ivanperez-keera:
Just to move things and close old issues:

If we move this into examples and add it to the travis script, we'll keep a good working example around until we can integrate it with everything else.

Can you do that?

turion:
I'd rather go back researching a little bit, if that's ok. I think I can remove the MSF constructor with [this function](https://github.com/ivanperez-keera/dunai/blob/7578ec68b94c2d75d574c95689e0c5d5c80842e5/src/Control/Monad/Trans/MSF/GenLift.hs#L90) from branch `dev_general_lifting`. It's ok if we close this PR for now and I reopen it once the general lifting stuff is settled.

ivanperez-keera:
Ok.

PR #90, opened 2018-01-02
Test all examples with travis

### LINKED ISSUE #58: Test examples with travis

turion:
* [x] Tested all examples
* [x] Updated to be compatible with the newest version of `base`
* [x] Added all examples to `.travis.yml`
* [ ] Modified deploy condition to only upload dunai and bearriver

_Warning:_ Before merging this, someone needs to review and/or test whether the deploy condition in `.travis.yml` works correctly. I don't know how to test this without actually triggering a release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
 before_install:
   - cd ${PACKAGEDIR}
   - sudo apt-get update -qq
-  - sudo apt-get install -qq libsdl-gfx1.2-dev libwxgtk3.0-dev
+  - sudo apt-get install -qq libsdl-gfx1.2-dev libwxgtk3.0-dev libbluetooth-dev libcwiid-dev
 
 deploy:
   provider: hackage
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,5 +31,6 @@ deploy:
     secure: nfxBR6dV3B+qmQcxiJsmQMPrRUvUZRfSPmhFWwFIjJCfjJLTteQg9V0QI2IpssnNos9qQrCF9qxsvSGSliC9rANYRz8jGz4y5/8XmS0LE71BDTvB8WyzWhKGB9SXNN4xozEIvFDdpEopeK3uh7yf3GXY30zDNhYbZaj38eUF0/3WfxcBBgKDjPgnVShNgoXkkwI6Vz+e1M42/f0IoUVdXWikmCuczMoVYKIkawr+qjCKn9jzKzsdxVYgFtKwOx4wzXvncfvsi1H90pHt4RZi0ufS0qhFxq8Wylx8s1J4i1AuC/2iuAhdEOTidSXGUpi3hDfPhvq70+5vJr9RxgdRA5SkqXT/cck5cC21uOuIw1mGs2Zic7jD16MrpDmZEgZq5QpsK7Hwy5B3+YXZ7y7gG8nxobg8Fj49IbYTuigGrlXJumAXMtU1mspRfWthySKLLrVyCqn8Jow8DcjtmQDGRf4yY2EPngf8iJ7VmFaUpiCPZZ2aE4FxjSpqgJLFun2szLsNLNVFkUHjHF/NFJzW6i+evy/oJmvk+SmI0D/YunbUmlgrN2+sy5TpwWquN9I78SDaGpP8yNlR7OutU6DgLpu9BigLTQHKQr+tJz30c4em3w2Gl1F+QM8sIEEvafEX8+h6WqpKuHi8sQlJKJziAnoEfgDC83tKJJpMc9hj3kE=
   on:
     branch: master
-    condition: ${TRAVIS_HASKELL_VERSION}=8.2.1
-    condition: ""${PACKAGEDIR} =~ ^(\.|examples/bearriver)$"
+    condition:
+      - ${TRAVIS_HASKELL_VERSION}=8.2.1
+      - ${PACKAGEDIR} =~ ^(\.|examples/bearriver)$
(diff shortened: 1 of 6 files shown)


turion:
Currently, the 7.6.3 cases fail because the version of bearriver on hackage requires `base>=4.7`. This is already fixed in `develop`, but not released yet.

turion:
* Some examples (`bouncingball`, `haskanoid`, ...) require external libraries. Those add 
* There seems to be a [bug](https://sourceforge.net/p/wxhaskell/bugs/116/) in `wxcore` which prevents [some GHC 8.2.1 jobs](https://travis-ci.org/ivanperez-keera/dunai/jobs/326844410) from completing.
* There is a strange error `configure: error: C compiler cannot create executables` (travis bug?) which prevents [haskanoid](https://travis-ci.org/ivanperez-keera/dunai/jobs/326821294) from completing.

ivanperez-keera:
>    Some examples (bouncingball, haskanoid, ...) require external libraries. Those add about 10 seconds to each job.

Does this change anything, regarding any decisions to be made wrt to accepting or rejecting the PR?

>    There seems to be a bug in wxcore which prevents discrete and keerahails on GHC 8.2.1 from completing.

Has this bug been reported to wxcore? Can this case be disabled so that it does not cause the travis job to fail.

>    There is a strange error configure: error: C compiler cannot create executables (travis bug?) which prevents haskanoid from completing.

Unless haskanoid is not compiled with the rest of the examples, this breaks the travis job. It either needs to be fixed (here or in the haskanoid project), or disabled (if it currently is already disabled, which I guess is the case).

turion:
>>    Some examples (bouncingball, haskanoid, ...) require external libraries. Those add about 10 seconds to each job.

> Does this change anything, regarding any decisions to be made wrt to accepting or rejecting the PR?

No, unless you want to optimise the duration of a travis run.

>>    There seems to be a bug in wxcore which prevents discrete and keerahails on GHC 8.2.1 from completing.

> Has this bug been reported to wxcore? Can this case be disabled so that it does not cause the travis job to fail.

Yes, I reported it: https://sourceforge.net/p/wxhaskell/bugs/116/
I believe this case could be disabled with a travis build matrix condition, but I'm not entirely sure how.

>>    There is a strange error configure: error: C compiler cannot create executables (travis bug?) which prevents haskanoid from completing.

> Unless haskanoid is not compiled with the rest of the examples, this breaks the travis job. It either needs to be fixed (here or in the haskanoid project), or disabled (if it currently is already disabled, which I guess is the case).

I don't understand. (How) Does haskanoid depend on other examples?

ivanperez-keera:
Currently, several examples are broken. See: https://travis-ci.org/ivanperez-keera/dunai/builds/326844429

ivanperez-keera:
>     Unless haskanoid is not compiled with the rest of the examples, this breaks the travis job.
> 
> I don't understand. (How) Does haskanoid depend on other examples?
> 

It doesn't, but it is included in the tests. It needs to be fixed or that test disabled.

turion:
I don't know how to fix any of the remaining problems. Of course, we can always take them out of the build matrix, but is that acceptable?

ivanperez-keera:
Some packages are not compiling due to their version boundaries in their respective cabal files. I'm not sure those are there because they need to, or because they were the default upon `cabal init`.

Those issues can be fixed. For everything else, I'm happy disabling some tests for now and revisiting this in the future.

This is already an improvement over what we had.

turion:
The version boundary issue is already fixed in `develop`. The reason it's not fixed on travis is because travis downloads dependencies from hackage, not from the repo. We could change this with adding sandboxes and add-source dependencies to the build process.

ivanperez-keera:
> The version boundary issue is already fixed in develop. 

Maybe for `bearriver`, but not for all the examples. Look:

```
$ grep -nHre '\<base\>'  --include='*.cabal' | column -t | grep -oe '^[^,]*' | sort -g
basic-examples/basic-examples.cabal:22:                   build-depends:  base  >=4.8  &&   <4.9
bearriver/bearriver.cabal:26:                             build-depends:  base  >=4.6  &&   <5
bouncingball/bearriver-examples-bouncingball.cabal:20:    build-depends:  base  >=4.6  &&   <      5
bouncingball/bearriver-examples-bouncingball.cabal:31:    build-depends:  base  >=4.6  &&   <      5
classicfrp/classicfrp.cabal:22:                           build-depends:  base  >=4.8  &&   <4.9
classicfrp/classicfrp.cabal:31:                           build-depends:  base  >=4.8  &&   <4.9
discrete/dunai-examples-discreterp-wxhaskell.cabal:14:    build-depends:  base  >=4.7  &&   <4.9
discrete/dunai-examples-discreterp-wxhaskell.cabal:19:    build-depends:  base  >=4.7  &&   <4.9
haskanoid/haskanoid.cabal:102:                            build-depends:  base  >=     4.6  &&     <             5
keerahails/dunai-examples-keerahails-wxhaskell.cabal:14:  build-depends:  base  >=4.7  &&   <4.9
list/bearriver-examples-bouncingball-list.cabal:19:       build-depends:  base  >=4.6  &&   <      5
paper/dunai-examples-paper.cabal:16:                      build-depends:  base  >=4.6  &&   <5
reversefrpzoo/wormholes/dunai-frp-wormholes.cabal:16:     build-depends:  base  >=4.6  &&   <      5
taggingmonad/taggingmonad.cabal:23:                       build-depends:  base  >=4.8  &&   <4.9

```

turion:
Oh, my bad, sorry. What we want, according to [this table](https://ghc.haskell.org/trac/ghc/wiki/Commentary/Libraries/VersionHistory), is `base >= 4.6` as long as we're sticking to GHC 7.6.3.

PR #91, opened 2018-01-09
Wormholes

turion:


ivanperez-keera:
This is neat. But should it be in `examples/` for now?

Presumably, we'd make one like this for each frp library in the reverse zoo, in which case, does it make sense to add all to the main library?

Another way of asking the same thing is: is this an elementary addition to dunai **and** is it expressed at the same level of abstraction as other dunai constructs, or is it an addition on top that should be in a library that imports dunai (in `examples/` or elsewhere).

turion:
I think this is a useful feature that should go in dunai directly. For example when you need to use dunai with multithreading (i.e. OpenGL), you'll probably need wormholes.

It could in principle be implemented as a library on top of dunai, but it would be tiny and harder to find. (More a documentation problem as long as we don't have a separate guide to dunai.)

PR #110, opened 2018-06-22
add is dead

ggreif:
First I wanted to say
``` haskell
add = arr $ uncurry (+)
```
but then I realised that it is not used anywhere.

### DIFF
--- a/examples/basic-examples/BasicExamples.hs
+++ b/examples/basic-examples/BasicExamples.hs
@@ -4,10 +4,6 @@ import Data.MonadicStreamFunction
 
 import Text.Read (readMaybe)
 
-
-add :: (Num n, Monad m) => MSF m (n, n) n
-add = arr (\(x, y) -> x + y)
-
 testSerial :: MSF IO () ()
 testSerial =   liftS (\_ -> getLine)
            >>> (arr id &&& arr reverse)


turion:
What exactly is the problem? Your refactoring is actually quite nice. This file contains examples of what we can do with `MSF`s, they don't necessarily need to be reused anywhere.

ivanperez-keera:
FYI, there is `paper/` folder that currently includes some examples we put in the paper, and the goal would be that it contained all of them.

ggreif:
@turion There is no problem, other than `add` is defined, but never used. So it is redundant. My patch removes it.

ivanperez-keera:
But that module exports everything, doesn't it?

ivanperez-keera:
@turion Is the module meant to be run as an executable, or from ghci?

turion:
I don't know anymore. I think that package is not important at all. It's not well-documented and doesn't have a lot of interesting features. As for basic examples, I think the ones from the article and elsewhere (to some extent, Rhine examples are also Dunai examples), are already sufficient, so maybe we should just delete the package. If there is anything important in it, we can maybe incorporate it in another well-documented example?

ivanperez-keera:
The whole `basic-examples` folder can go.

PR #111, opened 2018-06-22
More recent GHCs also work

ggreif:
I have even compiled it with GHC HEAD (v8.7)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ $ cabal install dunai
 
 ## Dependencies
 
-Dunai currently support GHC versions 7.6.3 to 8.2.1.
+Dunai currently support GHC versions 7.6.3 to 8.4.3.
 
 # Examples
 


turion:
That's great to know! It's also about the continuous integration, so I guess we won't claim that it runs on 8.4 before travis builds it (and all examples) successfully on 8.4.

ivanperez-keera:
I think a big issue at this point is that we have too many examples, so it takes forever to build. Adding a new build on travis for everything means maybe 1h of workers on travis' systems.

Also, the fact that an example does not build does not mean that dunai should not work. An example may not build because a library needed by the example only works with older versions of GHC, but dunai itself works with newer versions.

So, I am in principle not against increasing the base version boundaries of dunai, so long as everything in dunai's cabal file compiles with those versions of GHC on travis.

turion:
There are two things we could do, if we want to increase version boundaries (which I think is a good idea):

1. Group similar examples together in one cabal project. What takes longest on travis is the install phase (installing dependencies), so if we have fewer cabal projects in total, the same amount of code is checked in less time.
2. Put some, many, or all examples in a separate repository. Then travis is only run when we update the examples, or bump the dunai version used in the example. This also solves the problem that dunai is displayed as "build error" when it actually works, but some example has a broken dependency (the problem you just mentioned).

If you think that either or both is a good idea, I'm happy to send a PR.

PR #112, opened 2018-06-24
Adds edge detectors

turion:
I know this won't be merged soon. It mainly exists to close https://github.com/turion/rhine/issues/80.


ivanperez-keera:
This could perfectly be in sonnendemo for now. The fact that conceptually the might belong in dunai does not mean that it should be a priority to add them now.

There's been an issue open in dunai for a while, because it's important, and a priority. We spent too much time discussing small stuff and that made it impossible to discuss bigger stuff. And that issue got stuck.

I don't mind this being here, but don't complain if it does not get merged. Consider keeping this locally in your project. It's ok to at least ask if it can be merged, but do understand that you cannot complain if it's not given priority over other tasks.

I just came back from a trip to present the work in dunai and Rhine, after a lot time spent to create that work. It's not nice to suddenly complain because you don't see commits when this does not prevent you from extending dunai locally (like we all do with all libraries in real-world projects) and people are working on this project.

Please, be mindful of other people.

PR #113, opened 2018-06-27
Suggest three utilities for sources and sinks

ggreif:
With this you can obtain a network listener as
``` haskell
acceptor :: Socket -> MStream IO (Socket, SockAddr)
acceptor = stream . accept
```

### DIFF
--- a/src/Data/MonadicStreamFunction/Util.hs
+++ b/src/Data/MonadicStreamFunction/Util.hs
@@ -21,9 +21,20 @@ import Data.VectorSpace
 -- It can obtain the values from a monadic context.
 type MStream m a = MSF m () a
 
+-- | Obtain a stream by repetitively running a monadic action.
+stream :: Monad m => m a -> MStream m a
+stream act = MSF go
+  where go _ = do a <- act
+                  pure (a, stream act)
+
 -- | A sink is an 'MSF' that consumes inputs, while producing no output.
 -- It can consume the values with side effects.
 type MSink   m a = MSF m a ()
+
+-- | Obtain a sink by repetitively feeding a kleisli arrow.
+sink :: Monad m => (a -> m b) -> MSink m a
+sink kl = MSF go
+  where go a = kl a >> pure ((), sink kl)
 
 -- * Lifting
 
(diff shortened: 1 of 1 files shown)


ggreif:
Turns out, that further
``` haskell
sink = arrM . fmap void
```
but then one needs to
``` haskell
import Control.Monad (void)
```

src/Data/MonadicStreamFunction/Util.hs
```diff
@@ -67,6 +67,17 @@ mapMSF = MSF . consume
       (bs, sf'') <- consume sf' as
       b `seq` return (b:bs, sf'')
 
+-- | Apply an 'MSF' to every input, discarding results.
+-- Used for side-effects only.
+mapMSF_ :: Monad m => MSF m a b -> MSink m [a]
```
turion: That's maybe useful, but can't you rewrite it in terms of `mapMSF`? Something like `mapMSF . (>>> arr void)`. (I don't think performance will suffer, because laziness.) Using the `MSF` constructor is not encouraged.

ggreif:
Unfortunately performance may suffer because of `seq` in `mapMSF`, but I
did not check that closely.

Em sex, 29 de jun de 2018 às 18:41, Manuel Bärenz <notifications@github.com>
escreveu:

turion:
Ah, true. Good spot. Still I think there should be the more general lazy variant of `mapMSF`, and `mapMSF_` a special case of this. But I think @ivanperez-keera doesn't want to go into issues of laziness right now.

ivanperez-keera:
There is a much bigger plan for both Yampa and dunai, and no, performance is not considered a priority until more crucial pieces are in place for both. This is not dunai specific: performance adjustments generally require specific code adaptations that break abstractions and make code harder to maintain and rewrite. This includes issues of laziness, so long as they do not cause inherent leaks or prevent us from achieving the functional purpose. Note that extra strictness can lead to bottoms where laziness would not and, when that happens, it may be justified to add a specific function that behaves differently.

Wrt. to this issue, unfortunately, because these can written in terms of existing functions, I don't see the use, unless you can show that you can't do what you want with `mapMSF`.

ivanperez-keera:
If there is no reply, I'm going to close this. In general, dunai is extensible, and most non-trivial projects extend existing libraries, so consider adding it locally to your project. If the use is common in many projects, or if we conclude that it just presents the right level of abstraction, we can re-evaluate this in the future.

If `mapMSF_` is needed and cannot be expressed in terms of other combinators, please, try to show why and send just that.

PR #116, opened 2018-10-04
Removes incomplete patterns in Trans.MSF.Except

### LINKED ISSUE #115: -XNoMonadFailDesugaring
The version on hackage should be updated to add `-XNoMonadFailDesugaring` in the `.cabal` file. Otherwise GHC v8.6 will error out. Stackage nightly also now incorporates GHC-8.6, so this is getting relevant.

turion:
Hopefully fixes #115. @ggreif, does that work for you?

### DIFF
--- a/src/Control/Monad/Trans/MSF/Except.hs
+++ b/src/Control/Monad/Trans/MSF/Except.hs
@@ -18,6 +18,7 @@ import           Control.Monad              (liftM, ap)
 import           Control.Monad.Trans.Class
 import           Control.Monad.Trans.Except hiding (liftCallCC, liftListen, liftPass) -- Avoid conflicting exports
 import           Control.Monad.Trans.Maybe
+import           Data.Either                (fromLeft, fromRight)
 
 -- Internal
 -- import Control.Monad.Trans.MSF.GenLift
@@ -175,7 +176,8 @@ safely :: Monad m => MSFExcept m a b Empty -> MSF m a b
 safely (MSFExcept msf) = safely' msf
   where
     safely' msf = MSF $ \a -> do
-      Right (b, msf') <- runExceptT $ unMSF msf a
+      (b, msf') <- fromRight (error "safely: Received `Left`")
+        <$> (runExceptT $ unMSF msf a)
       return (b, safely' msf')
 
 -- | An 'MSF' without an 'ExceptT' layer never throws an exception,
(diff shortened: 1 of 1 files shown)


ggreif:
Looking at the travis-ci (GHC 8.0.2) logs:

```
[17 of 22] Compiling Control.Monad.Trans.MSF.Except ( src/Control/Monad/Trans/MSF/Except.hs, dist/build/Control/Monad/Trans/MSF/Except.o )
src/Control/Monad/Trans/MSF/Except.hs:21:46: error:
    Module ‘Data.Either’ does not export ‘fromLeft’
src/Control/Monad/Trans/MSF/Except.hs:21:56: error:
    Module ‘Data.Either’ does not export ‘fromRight’
```

turion:
Good that we have Travis. Bad that we still support so old GHC versions. @ivanperez-keera, rather CPP `#ifdef` or defining those two functions independently of base?

ivanperez-keera:
Come on, 8.0 is far from old.

You should be CPP-ed.

turion:
Ok, but this next commit is really just to satisfy stone-age GHCs ;) can't test locally because NixOS doesn't support pre 7.10 GHCs.

ivanperez-keera:
Thanks.

But I see that you are still pushing to a PR. If this is not ready and more needs to be discussed, please, instead of sending a PR, discuss in an issue, implement and check locally or in a branch, and send only when it's clear and finished. Also, generally, consider rebasing the changes in the branch to collapse them or reorganize them so that's it's easier to understand what the change does.

ivanperez-keera:
Also, why use Monad if you can get away with just applicative or functor (if you can, I'm not sure)?

turion:
Yes, this chain of commits is a bit of a mess. The reason why it's so clunky is that I don't even have a development environment for that old GHCs. 

I've used `Monad` here because that's going to be required anyways (`Monad m => Monad (MSFExcept m a b)`). I've not used `Functor` because that would infect any other place that uses these, and would require a lot of changes. But I can change that if you want.

ivanperez-keera:
I see. It's ok, I too need to commit many times with small fixes because of Travis.

However, while you are working on this, the right way of testing this is to either create a branch, or fork on github, commit to your fork, and send a clean PR when it's ready.

For info: GHC 7.6 is still present in stable versions of Debian and Ubuntu currently being supported by companies as part of their LTS programme. These happen to be distros used in production. I know, for example, that Keera's clients' work is partly maintained in a server in which only GHC 7.6 is directly available. Many libraries support these "old" versions of GHC, and many support at least 3 back from the current stable (which, until less than a year ago, included 7.10!!).

Please, consider investigating in a branch or fork, and then reorganizing and sending the PR.

turion:
Ok, noted for future workflows. Shall I rebase in some way and `git push --force` or do you prefer to squash from the github UI?

ivanperez-keera:
I don't know how to "squash from the github UI".

If it affects only your branch, you can `push --force`.

turion:
Ah, if you click on the triangle button right next to "Merge pull request", you can select "Squash and merge". I believe you can then combine all commits to a single one and give that a new name. What do you prefer? (The rebase changes would only affect this branch, so `git push --force` would work as well)

PR #120, opened 2018-10-13
Effected GHC 8.6.1 Compatibility

0xCM:
* This was primarily an exercise in replacing Monad constraints with MonadFail constraints.
* Applied similar changes to Rhine


turion:
Thanks! But note that if #116 is going forward we won't even need `MonadFail`.

PR #128, opened 2018-11-25
Relax Type-Class Constraints

diesalbla:
The functions `morphGS` and `feedback` in the `InternalCore.hs` and `Core.hs` files were constrained to `Monad`. However, they only had a `do` with one bind followed by `return`, which is a `fmap`. We use this to rewrite these functions and relax their type-class constraints from `Monad` to `Functor`.

### DIFF
--- a/src/Data/MonadicStreamFunction/Core.hs
+++ b/src/Data/MonadicStreamFunction/Core.hs
@@ -63,7 +63,7 @@ import Control.Monad.Trans.Class
 import Data.Tuple (swap)
 import Prelude hiding ((.), id, sum)
 
-import Data.MonadicStreamFunction.InternalCore (MSF, morphGS, feedback, reactimate, embed)
+import Data.MonadicStreamFunction.InternalCore (MSF, morphGS, feedback, reactimate, embed, inner_arrM)
 
 -- * Definitions
 
@@ -100,28 +100,24 @@ instance (Functor m, Monad m) => Applicative (MSF m a) where
 -- ** Lifting point-wise computations
 
 -- | Lifts a monadic computation into a Stream.
-constM :: Monad m => m b -> MSF m a b
+constM :: Functor m => m b -> MSF m a b
 constM = arrM . const
 
 -- | Apply a monadic transformation to every element of the input stream.
 --
 -- Generalisation of 'arr' from 'Arrow' to monadic functions.
-arrM :: Monad m => (a -> m b) -> MSF m a b
---arrM f = go
---  where go = MSF $ \a -> do
---               b <- f a
---               return (b, go)
-arrM f = morphGS (\i a -> i a >>= \(_,c) -> f a >>= \b -> return (b, c)) C.id
+arrM :: Functor m => (a -> m b) -> MSF m a b
+arrM = inner_arrM
 
 -- | Monadic lifting from one monad into another
-liftBaseM :: (Monad m2, MonadBase m1 m2) => (a -> m1 b) -> MSF m2 a b
+liftBaseM :: (Functor m2, MonadBase m1 m2) => (a -> m1 b) -> MSF m2 a b
 liftBaseM = arrM . (liftBase .)
 
 -- ** MSF combinators that apply monad transformations
 
 -- | Lift innermost monadic actions in monad stack (generalisation of
 -- 'liftIO').
-liftBaseS :: (Monad m2, MonadBase m1 m2) => MSF m1 a b -> MSF m2 a b
+liftBaseS :: (Functor m2, MonadBase m1 m2) => MSF m1 a b -> MSF m2 a b
 liftBaseS = morphS liftBase
 
 -- *** MonadBase
@@ -139,7 +135,7 @@ sf1 >>>^ sf2 = sf1 >>> liftBaseS sf2
 
 -- | Lift inner monadic actions in monad stacks.
 
-liftTransS :: (MonadTrans t, Monad m, Monad (t m))
+liftTransS :: (MonadTrans t, Monad m, Functor (t m))
            => MSF m a b
            -> MSF (t m) a b
 liftTransS = morphS lift
@@ -150,7 +146,7 @@ liftTransS = morphS lift
 --
 -- This is just a convenience function when you have a function to move across
 -- monads, because the signature of 'morphGS' is a bit complex.
-morphS :: (Monad m2, Monad m1)
+morphS :: (Functor m2, Functor m1)
       => (forall c . m1 c -> m2 c)
       -> MSF m1 a b
       -> MSF m2 a b
--- a/src/Data/MonadicStreamFunction/InternalCore.hs
+++ b/src/Data/MonadicStreamFunction/InternalCore.hs
@@ -88,23 +88,20 @@ instance Monad m => Category (MSF m) where
 -- types, and doesn't influence control flow.
 -- Other handling functions like exception handling or 'ListT' broadcasting
 -- necessarily change control flow.
-morphGS :: Monad m2
+morphGS :: Functor m2
         => (forall c . (a1 -> m1 (b1, c)) -> (a2 -> m2 (b2, c)))
           -- ^ The natural transformation. @mi@, @ai@ and @bi@ for @i = 1, 2@
           --   can be chosen freely, but @c@ must be universally quantified
         -> MSF m1 a1 b1
         -> MSF m2 a2 b2
-morphGS morph msf = MSF $ \a2 -> do
-  (b2, msf') <- morph (unMSF msf) a2
-  return (b2, morphGS morph msf')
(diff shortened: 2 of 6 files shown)


diesalbla:
> I understand the reason to do this in abstract terms but, in practice, have you found a functor you wanted to use that wasn't a monad?

I have not tried to use it, so, so far, no. I was looking at the code and noticed that the constraints could be lowered (or raised). I assume that you only need to tell in _each_  function what constraints apply to it; and let the compiler figure out how to solve them.

ivanperez-keera:
I've played around with it a bit. My concern is that it makes the code of the morphing functions (and the inner_arrM) harder to understand.

ivanperez-keera:
>  IMHO, one only needs to tell in each function what constraints apply to it, and let the compiler figure out how to mix them.

Sure, I suspect that's what we all do.

diesalbla:
> I've played around with it a bit. My concern is that it makes the code of the morphing functions (and the inner_arrM) harder to understand.

I will try to clean it up a bit tomorrow, to see if it can be made more readable.

src/Data/MonadicStreamFunction/Core.hs
```diff
@@ -149,27 +131,9 @@ liftTransS = morphS lift
 -- | Apply trans-monadic actions (in an arbitrary way).
 --
 -- This is just a convenience function when you have a function to move across
--- monads, because the signature of 'morphGS' is a bit complex.
-morphS :: (Monad m2, Monad m1)
-      => (forall c . m1 c -> m2 c)
-      -> MSF m1 a b
-      -> MSF m2 a b
-morphS morph = morphGS morph'
-  where
```
diesalbla: Since `morphS` is no longer uses the very generic `morphGS`, the explanation below is no longer unnecessary.

src/Data/MonadicStreamFunction/InternalCore.hs
```diff
@@ -88,23 +87,23 @@ instance Monad m => Category (MSF m) where
 -- types, and doesn't influence control flow.
 -- Other handling functions like exception handling or 'ListT' broadcasting
 -- necessarily change control flow.
-morphGS :: Monad m2
-        => (forall c . (a1 -> m1 (b1, c)) -> (a2 -> m2 (b2, c)))
+morphGS :: Functor m2
+        => (InMorph m1 a1 b1 ~> InMorph m2 a2 b2)
```
turion: I personally find this type synonym no readability improvement.

src/Data/MonadicStreamFunction/InternalCore.hs
```diff
@@ -133,3 +132,66 @@ reactimate :: Monad m => MSF m () () -> m ()
 reactimate sf = do
   (_, sf') <- unMSF sf ()
   reactimate sf'
+
+-- Implementation of some common operations for type-class instances, without the instances
+map_msf :: Functor m => (b -> c) -> MSF m a b -> MSF m a c
```
turion: Underscores in names are not OK.

turion: This should rather be a `Functor` instance for `MSF`s, and not a standalone function.

src/Data/MonadicStreamFunction/InternalCore.hs
```diff
@@ -133,3 +132,66 @@ reactimate :: Monad m => MSF m () () -> m ()
...
+map_msf f sf = MSF $ \ a -> fS <$> unMSF sf a where
```
turion: An implementation without the `MSF` constructor would be preferable. If you want one with the `MSF` constructor, have a look at https://github.com/ivanperez-keera/dunai/blob/c74aebe596791651edba543998ec85328a58b2e7/src/Data/MonadicStreamFunction/Core.hs#L89, it's already there in a comment.

turion:
I think that relaxing to `Functor` is an issue with `do`-notation, not with `dunai`. You know that for `Applicative`s, you can use `do`-notation with the extension `ApplicativeDo`. What's missing in GHC is a "`FunctorDo`"-notation which translates any `do`-block with a single statement into an application of `fmap`. (I asked SPJ about such a feature once, but he thought that it's not as important.) So my opinion is that this ought to be a GHC ticket, and individual libraries shouldn't need to work around this.

Besides, it's not clear to me what the benefit is. Yes, sometimes `Functor` is more efficient than `Monad`, but in the end you're going to need `Monad` anyways for any nontrivial program, since `reactimate` needs it. So at most you're going to win a tiny bit of efficiency, but it's not clear whether premature optimisation is worth it here.

src/Data/MonadicStreamFunction/InternalCore.hs
diesalbla: Yeah, that was the one I took as a base.

src/Data/MonadicStreamFunction/InternalCore.hs
diesalbla: I thought of that, but I did not want to add any type-class instances, since they impact modularity. A function can be imported manually.

diesalbla:
> Besides, it's not clear to me what the benefit is. Yes, sometimes Functor is more efficient than Monad, but in the end you're going to need Monad anyways for any nontrivial program, since reactimate needs it. So at most you're going to win a tiny bit of efficiency, but it's not clear whether premature optimisation is worth it here.

It was not at all about efficiency, just simplicity in the sense of minimal constraints. I was a bit confused at first when I saw the `Monad` constraints. 

> What's missing in GHC is a "FunctorDo"-notation which translates any do-block with a single statement into an application of fmap. 

That looks a great idea! However, even if the compiler can infer a stronger constraint (`Functor`), it could not replace the one in the source. Besides, I could never be as quick to turn a `do-return` into an`fmap` as the compiler may be. 

Readability may depend on formatting and operators. Using `<&>` operator, you may turn one code into the other: 
```Haskell
feedback :: Monad m => c -> MSF m (a, c) (b, c) -> MSF m a b
feedback c sf = MSF $ \a -> do
  ((b', c'), sf') <- unMSF sf (a, c)
  return (b', feedback c' sf') 

feedback :: Functor m => c -> MSF m (a, c) (b, c) -> MSF m a b
feedback c sf = MSF $ \a -> 
    unMSF sf (a, c)  <&>  
       \ ((b', c'), sf') -> (b', feedback c' sf') 
```
This two blocks of code  look fairly similar. Here, the `<&>` would be similar to the `>>=`. 

>  Thanks a lot for the effort you put into this. But I think that it's mostly not worth it, and quite a few parts of the code get harder to read. We should use the MSF constructor in as few places as possible.

One problem I've found is that [GHC versions before `7.10`](https://wiki.haskell.org/Functor-Applicative-Monad_Proposal) do not infer `Functor` directly from `Monad`, so several functions needed to extend those constraints. This, and the issues you mentioned, made the PR a bit less useful than I thought at first.

src/Data/MonadicStreamFunction/InternalCore.hs
turion: I meant both, although I personally quite like `~>`. (But it's @ivanperez-keera you need to convince ;) I'm just saying my personal opinion) If `InMorph` is used elsewhere, it would be good style to introduce this synonym, but I don't think it's needed much.

turion:
> > What's missing in GHC is a "FunctorDo"-notation which translates any do-block with a single statement into an application of fmap.
> 
> That would be a great improvement, to be sure. However, even if the compiler can infer a stronger constraint, it could not replace the one in the source.
Yes, but we could then write
```haskell
{-# LANGUAGE FunctorDo #-}
feedback :: Functor m => c -> MSF m (a, c) (b, c) -> MSF m a b
feedback c sf = MSF $ \a -> do
  ((b', c'), sf') <- unMSF sf (a, c)
  return (b', feedback c' sf')
```
The best of two worlds.

> Readability may depend on formatting and operators. Using `<&>` operator, you may turn one code into the other:
> [...]
I personally find it ugly and hard to read, to be honest.

> One problem I've found is that [GHC versions before `7.10`](https://wiki.haskell.org/Functor-Applicative-Monad_Proposal) do not infer `Functor` directly from `Monad`, so several functions needed to extend those constraints. This, and the issues you mentioned, made the PR a bit less useful than I thought at first.
Yes, that is a nuisance, but one day we'll be rid of older GHCs. Just some patience...

diesalbla:
> We should use the MSF constructor in as few places as possible.

I understand that you may want to hide the datatype behind a few choicy constructors, particularly the ones from type-class instances, and build on those. The problem I saw was that most of the implementations were based on `>>>`, or others, that were forcing a Monad constraint. That is why I added some of those `_msf` functions in `InternalCore`, to by-pass that.

In any case, I am not too involved with `dunai` itself; this is something I could offer for your convenience. Let me know if there is some part that may be of use, or if it is too far off your track.

Issue #134, opened 2019-02-10
Doc: Link to Yampa as related project from README


ivanperez-keera:
Yampa could also be mentioned in the README as a related project, since it's what Dunai emerged from.

Issue #133, opened 2019-02-10
Link to Rhine paper and Rhine project from Readme


ivanperez-keera:
The Rhine project by @turion complements Dunai with asynchronicity, and it should be mentioned in the README as a related project.

PR #160, opened 2019-10-01
Added Related Projects Section

dynomite567:
Solves issues #133 and #134 by adding a Related projects section in the README.md file

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,10 @@ You may have to use `cabal sandbox add-source` and install all libs with one
 `cabal install` invocation to link against the same versions of all common
 dependencies.
 
+# Related Projects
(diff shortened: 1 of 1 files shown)


ivanperez-keera:
Thanks, @dynomite567 . There's also a "Rhine: FRP with type-level clocks" paper that could be included. Can you add a commit with that as well as part of #133 ?

dynomite567:
Just want to make sure before I add it, this is it right?
https://dl.acm.org/citation.cfm?id=3242757

dynomite567:
Done. Also rebased with latest commits to avoid conflicts.

ivanperez-keera:
Thanks!

Issue #149, opened 2019-09-29
Bearriver: define accumBy


ivanperez-keera:
The function `accumBy` is pretty useful, but missing (we have `accumHoldBy`).

It would be useful to include it.

ivanperez-keera:
Subtask of #40 .

chriz-keera:
Hi,

could this be a solution for `accumBy`:

```
accumBy :: Monad m => (b -> a -> b) -> b -> SF m (Event a) (Event b)
accumBy f b = feedback NoEvent $ arr $ \(a, b') ->
  let eB = (fmap f b') <*> a
  in let b'' = if isEvent eB then eB else b'
     in (b'', b'')
```
And if so, is there a way to put the two lets in one?

turion:
I'd rather use `accumulateWith` and `mapMaybeS` from `dunai` and pre- & postcompose with the isomorphism `Maybe a = Event a`. In fact, there could be a function `mapEventS :: SF m a b -> SF m (Event a) (Event b)`.

chriz-keera:
What do you think about this?

```
accumBy :: Monad m => (b -> a -> b) -> b -> SF m (Event a) (Event b)
accumBy f b = mapEventS (accumulateWith (flip f) b)                                        
  where
    mapEventS :: Monad m => MSF m a b -> MSF m (Event a) (Event b)
    mapEventS msf = proc eventA -> case eventA of
      Event a -> arr Event <<< msf -< a
      NoEvent -> returnA           -< NoEvent
```

turion:
Sounds good to me in principle. I'd make mapEventS top level though, document it, and use $ in the definition of accumBy.
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

chriz-keera:
Thanks for the help!

PR #164, opened 2019-10-04
Implements accumBy and mapEventS. Refs #149.

chriz-keera:


ivanperez-keera:
@turion Is this what you had in mind?

ivanperez-keera:
(Never mind. I saw your comments on the issue.)

Issue #165, opened 2019-10-10
Help setting up repository for cabal user by adding cabal.project


Rizary:
Since I am a cabal user, I need to add `cabal.project` to start working on #96. 

So I start this issue, and add PR separately to add `cabal.project`. Cabal is already version 3.0 and by only using `cabal build` we can build the project easily. However, I know that several projects in the folder `examples` can't be built. But most of them not because of `cabal` behavior.

Adding `cabal.project` file to your repo helped cabal's user a lot.

ivanperez-keera:
I'm happy supporting newer versions of cabal, but I see these packages as separate (but related).

Doing this right suggests that we re-organize stuff.

As a first attempt, I've put all the files into subdirs.

https://github.com/ivanperez-keera/dunai/tree/develop-reorg

@turion Would you find a repo organized this way easier to navigate? Do you think it's an improvement?

turion:
I find that a great organisation! It needs some tweaks in `.travis.yml` still.

ivanperez-keera:
Correct, and a lot of the README.md might have to come to the top level. There would be some duplucation there, unless we can just link to the one in `dunai/`.

ivanperez-keera:
You can just link to it: https://github.com/ivanperez-keera/dunai/tree/develop-test-readme

ivanperez-keera:
There's some bug on github (just reported it). Your last comment "Ah, yes. The issue there..." should come before my last comment "You can just link to it".

turion:
> You can just link to it: https://github.com/ivanperez-keera/dunai/tree/develop-test-readme

What kind of link is that? A Linux symlink?

turion:
Ah, yes. The issue there is that the top level readme is displayed on github, but the readme in `dunai/` is displayed on hackage. Ideally, those two should agree, but I don't know how to do that reasonably. In `rhine`, I opted for a hyperlink on hackage to github..

ivanperez-keera:
Yes
.

PR #166, opened 2019-10-10
adding cabal.project to help cabal user get started

Rizary:
As mentioned in #165

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,0 +1,12 @@
+packages:
+  ./.
+  examples/bearriver/
+  examples/bouncingball/
+  examples/classicfrp/
+--  examples/discrete/
+  examples/haskanoid/
+--  examples/keerahails/
+  examples/list/
+--  examples/paper/
+  examples/reversefrpzoo/wormholes/
+  examples/taggingmonad/
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@ cabal-dev
 *.dyn_hi
 .hpc
 .hsenv
+.ghc.*
+cabal.project.local
 .cabal-sandbox/
 cabal.sandbox.config
 *.prof


cabal.project
```diff
@@ -0,0 +1,13 @@
+with-compiler: ghc-8.6.3
```
ivanperez-keera: Why does this file include a particular compiler version? What happens if users do not have that version?

cabal.project
```diff
@@ -0,0 +1,13 @@
...
+packages:
+  dunai.cabal
```
ivanperez-keera: I'm a bit confused by the fact that this list includes pointers to a cabal file and several dirs. Is it possible to just say "." instead of `dunai.cabal`?

cabal.project
Rizary: if I commented on the `with-compiler` it will use GHC PATH.

cabal.project
Rizary: the several dirs is needed so user can run `cabal build all` (for cabal version 3.0.0.0) to make cabal build all package contained in the examples folder.

it can be `./.` so it will find `.cabal` within root directory.

ivanperez-keera:
I don't think this file should mention a specific GHC version. It is the user's responsibility to determine which version to use, either with their `$PATH` or with additional `cabal` flags.

cabal.project
ivanperez-keera: Without being a cabal 3.0 expert, that seems to make more sense to me.

Rizary:
I've commented that out so user can be aware of `with-compiler` option.

cabal.project
Rizary: it is already happened in cabal new-build style (around version 2.4)

ivanperez-keera:
Thanks.

I think it's best to remove that line altogether. I know I may seem pretty annoying about this, but I generally think it's best not to include commented options or code in projects.

The place to check for other options is the cabal documentation (and the place for old/commented code is, generally, the repo history).

cabal.project
ivanperez-keera: True. Although I never truly updated any projects to work with `cabal new-build` style in 2.4, or with any additions in 3.0. I tend to purposefully delay adopting new versions of cabal & GHC, or new and fancy tools :)

Rizary:
Agree on that. So, I already delete the commented options.

cabal.project
```diff
@@ -0,0 +1,12 @@
+packages:
+  ./.
+  examples/bearriver/
+  examples/bouncingball/
+  examples/classicfrp/
+--  examples/discrete/
```
turion, line 6: Why are several examples commented?

cabal.project
```diff
@@ -0,0 +1,12 @@
+packages:
+  ./.
+  examples/bearriver/
```
turion, line 3: Wouldn't it be better to use a wildcard `examples/*/*.cabal examples/*/*/*.cabal`? That way, new examples get automatically CI'd.

cabal.project
```diff
@@ -0,0 +1,12 @@
...
+  examples/bouncingball/
+  examples/classicfrp/
+--  examples/discrete/
+  examples/haskanoid/
+--  examples/keerahails/
+  examples/list/
+--  examples/paper/
+  examples/reversefrpzoo/wormholes/
+  examples/taggingmonad/
```
turion, line 12: Missing a newline.

ivanperez-keera:
@Rizary A version of this change has been merged into dunai now. I am closing this one, but you and @turion will both be credited for the contribution.

I normally do not use cabal.project / cabal new-build, so it would be really helpful if you can send a PR if you see that anything regarding the `new-build` is broken.

turion:
That's awesome! I often use cabal projects, so I'll probably notice.

Issue #156, opened 2019-09-30
Link to talks on FRP and MSF from Scala perspective


ivanperez-keera:
Someone recently gave a talk about Monadic Stream Functions and an implementation in Scala.

It would be useful to link to their project from our README.

KatSaldivar:
Hi! Can you clarify what project are you are referring to? Is it the Scala project or the project of this person who gave the talk? I did a search for Monadic Stream Functions in Scala and there were a few results but none that obviously stood out as being the project you were referring to.

ivanperez-keera:
Hi @KatSaldivar . Thanks for taking a look! :)

I was referring to this particular work:

https://twitter.com/typelevel/status/1142391027732320257

I knew there was the talk. 13minutes into the talk, @diesalbla presented Monadic Stream Functions, and I assumed there would be an implementation in scala somewhere. Perhaps the author of the talk can confirm.

I suggest:
- Introducing a section after 'Further reading' called 'See also'
- Putting a link to the talk within that section, with using the title for the link, followed by a very short summary (5-10 words).
- If there is an implementation, linking to it.

Have you found any other implementations, by the way?

ivanperez-keera:
Add links to also the following two talks by the same author, either as part of this issue or as a separate issue:

https://skillsmatter.com/skillscasts/12542-london-scala-users-group

https://skillsmatter.com/skillscasts/12819-introduction-to-functional-reactive-programming-in-scala

ivanperez-keera:
Skillsmatter is, at least temporarily, closed, and links no longer work. Closed with the link provided by @Rizary .

PR #167, opened 2019-10-22
adding video reference from typelevel conference

Rizary:
As requested in https://github.com/ivanperez-keera/dunai/issues/156

I can't find any implementation mentioned in the slide.

I change the title name and put subsection to differentiate between reading and video.

For additional information, I just watched your talk about Rhine at ICFP 2018, and I think it suits also in dunai's video section. Do you think I should open another commit or put it in this commit?

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,9 @@ emosewa si lleksaH
 ^C
 ```
 
-# Further reading
+# Further references
+
+## Reading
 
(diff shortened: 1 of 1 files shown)


ivanperez-keera:
This was fixed and merged (crediting @Rizary ), so it does not show here as merged, but it was.

ivanperez-keera:
And thanks!

Issue #172, opened 2020-01-18
Remove vector definitions from bearriver


ivanperez-keera:
In a recent version, Yampa started using a separate package for vector and affine spaces. If these were used by bearriver, there would be less code to maintain.

If I am not mistaken, this is a matter of removing modules, adding a dependency in the cabal file, and making sure things still compile.

Issue #13, opened 2016-11-07
Use an external VectorSpace


turion:
Following the philosophy that dunai has to be slim, `VectorSpace` should really be a dependency. If there is no slim package supplying us with vector spaces, we should fork one.

This is not an API change or API addition, and its not urgent.

ivanperez-keera:
This is an API removal. It would impact the version major, and show go in release 0.3.

ivanperez-keera:
By the way, this implementation of [vector-space](https://hackage.haskell.org/package/vector-space), which I wanted to use for Yampa in the past, also depends on void.

Of course there's [linear](http://hackage.haskell.org/package/linear), but that just depends on half the universe.

Are there any simpler alternatives? If not, I'm happy externalising dunai's vector-space and moving forward with this.

Since I'm not an expert in vector spaces, can I ask: what are the key differences between the two? If any of those differences actually "define" or are fundamental to this implementation, then they could help find a suitable name for this library.

ivanperez-keera:
(Again, I'm happy depending on `vector-space`, so long as does the job, is well maintained, and does not depend on TH.)

turion:
I think `linear` is not the right package for us. `vector-space` sounds good. It depends on `template-haskell` at some point though: `vector-space -> void -> semigroups -> tagged -> template-haskell`.

Then I propose to fork `vector-space` without the dependency on `void`, and to use that in Dunai.

turion:
Looking into this further, [`void` doesn't even depend on semigroups for GHC >= 7.9](https://hackage.haskell.org/package/void-0.7.2/dependencies). So an even more elegant solution would be to deprecate GHC < 7.9 and use `vector-space` and `void`.

ivanperez-keera:
Ok. So, we can then delay this until 7.8 can be deprecated, and then this will naturally fall into place.

turion:
What is needed to deprecate 7.8? Is there anything I can do? (E.g. porting existing apps to 7.10?) Should we open a separate issue?

ivanperez-keera:
Unfortunately, no. But thanks. It's just a matter of patience.

ivanperez-keera:
Following a discussion in http://mailman.cs.yale.edu/pipermail/yampa-users/2018-October/000478.html, it seems like we cannot depend on an existing vector-space/linear for now, but we can externalize Yampa's, find the common factor with dunai, and at least not have it in Dunai/Yampa.

[...]

PR #173, opened 2020-01-20
Removed Yampa's affine space definitions; made Dunai's internal

Linearity:
These changes address issues #172 and #13, and they work for me. The "dunai" package no longer exposes `Data.VectorSpace`. Now my `import Data.VectorSpace` statements are no longer ambiguous between the "dunai" and "simple-affine-space" packages.

Even with this change, `FRP.BearRiver` still re-exports `Data.VectorSpace`. I assume this is just for convenience, but it forces anyone using another vector space class with the same method names to resolve name conflicts whenever she imports `FRP.BearRiver`. In particular, it's inconvenient to use the "linear" package, with its class `Additive` and methods like `(^+^)`, alongside with `FRP.BearRiver`. I'd prefer just to import `Data.VectorSpace` on my own.


ivanperez-keera:
@Linearity / Alex Thanks!

If I understand this change correctly, this would make it impossible to define your own instances of vector space to use with bearriver's integral, right?

This indeed hints to a bigger issue with vector spaces (see #13), and it would probably be best, just to move forward, to stick to simple-affine-space and add any `RModule` or more interesting type there, until we can switch to a different library. 

I'll write on top of your commit to eliminate the need for `RModule` and dunai's `Data.VectorSpace` module before I merge this into develop.

Also, there's some binary .swp file (vim?) that should have been ignored by git. Different issue, though.

ivanperez-keera:
@Linearity Does this work for you? https://github.com/ivanperez-keera/dunai/tree/Linearity-develop

ivanperez-keera:
@turion If anyone is using dunai's `Data.VectorSpace`, it's you. Is this going to break all of your code?

Linearity:
That branch does work for me. I don't understand those error messages from the continuous integration system. But the only checks I did were `stack build` and `stack build bearriver`, though, so maybe there's something else to check.

In my Yampa application I defined an orphan instance of `VectorSpace`, as defined in "simple-affine-space," for the "linear" package's `V2 Double` type that I'm using a lot. It allows me to run `integral` on signals of type `V2 Double`. The "Yampa" package does not provide the module `Data.VectorSpace`, so "dunai" not providing it either is OK. True, it does require me to use "simple-affine-space" knowing _a priori_ that Yampa uses the same package.

The `FRP.Yampa` module does re-export `Data.VectorSpace`, just as `FRP.BearRiver` does, but I removed that in my local version for the same reason I advocated removing it from `FRP.BearRiver` above.

Linearity:
> Also, there's some binary .swp file (vim?) that should have been ignored by git. Different issue, though.
Yeah, sorry. I've been using darcs so much lately that I forgot that `git add` is the same as both `darcs add` and `darcs record`. Thanks for fixing it.

ivanperez-keera:
So, this commit has been merged, but I put something on top and rebased. I'll close this, but you are still credited for your commit :) Thanks!

turion:
>  If anyone is using dunai's `Data.VectorSpace` it's you. Is this going to break all of your code?

Yes, but this refactoring is worth it :)

PR #177, opened 2020-01-25
bearriver: Remove extra time step during switches

Linearity:
When I ported some switching signal functions from Yampa to Bear River, I noticed what looked like dropped frames, that is, extra-long time steps.

I realized that `switch` erroneously represents two normal time steps in one: it runs both its first and its second signal function over the same length of time during the switching step. It skips the first output in the second signal function, for which the local time is supposed to be equal to zero.

My solution is to set the first time step of the second signal function to zero using

    local :: Monad m => (DTime -> DTime) -> ReaderT DTime m a -> ReaderT DTime m a

With this change my ported signal functions behave as they did in Yampa.

This change would invalidate any assumptions that the time step is nonzero. For example, `derivative` is just a difference quotient with the time step in the denominator. Its output will be `NaN` or `Infinity` if the step is equal to zero. Also, any signal function whose internal state changes even over a zero time step might skip its second output instead of its first.

I don't think these are big problems. State changes over zero time are bugs to begin with, since they are inconsistent with the continuous-time semantics of signal functions. I believe no primitive signal function in Bear River behaves this way. And Yampa has for a long time warned that `derivative` is risky:

> A very crude version of a derivative. It simply divides the value difference by the time difference. Use at your own risk.

I have never used `derivative`, and I would probably write my own version if the need arose. For convenience's sake Bear River could reasonably provide a `safeDerivative` that checks for the singularity.

For clarity I also slightly reorganized `reactimate` to call `MSF.switch` directly rather than binding an intermediate function to `switch` locally.

### DIFF
--- a/examples/bearriver/src/FRP/BearRiver.hs
+++ b/examples/bearriver/src/FRP/BearRiver.hs
@@ -409,14 +409,14 @@ switch :: Monad m => SF m a (b, Event c) -> (c -> SF m a b) -> SF m a b
 switch sf sfC = MSF $ \a -> do
   (o, ct) <- unMSF sf a
   case o of
-    (_, Event c) -> unMSF (sfC c) a
+    (_, Event c) -> local (const 0) (unMSF (sfC c) a)
     (b, NoEvent) -> return (b, switch ct sfC)
 
 dSwitch ::  Monad m => SF m a (b, Event c) -> (c -> SF m a b) -> SF m a b
 dSwitch sf sfC = MSF $ \a -> do
   (o, ct) <- unMSF sf a
   case o of
-    (b, Event c) -> do (_,ct') <- unMSF (sfC c) a
+    (b, Event c) -> do (_,ct') <- local (const 0) (unMSF (sfC c) a)
                        return (b, ct')
     (b, NoEvent) -> return (b, dSwitch ct sfC)
 
@@ -540,8 +540,8 @@ reactimate senseI sense actuate sf = do
  where sfIO        = morphS (return.runIdentity) (runReaderS sf)
 
        -- Sense
-       senseSF     = switch senseFirst senseRest
-       senseFirst  = constM senseI >>> (arr $ \x -> ((0, x), Event x))
+       senseSF     = MSF.switch senseFirst senseRest
+       senseFirst  = constM senseI >>> (arr $ \x -> ((0, x), Just x))
        senseRest a = constM (sense True) >>> (arr id *** keepLast a)
 
        keepLast :: Monad m => a -> MSF m (Maybe a) a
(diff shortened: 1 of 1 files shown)


ivanperez-keera:
I do not think this change is conceptually correct. If at all, the signal with the dt to zero should be the one switched on after the switch, not before.

Linearity:
This indeed makes `dt = 0` for the signal function _after_ the switch, not before.

In fact, already, the one active before the switch begins with `dt = 0` as well, at least for the outermost `switch` combination. This is clear in `reactimate`:

    senseFirst  = constM senseI >>> (arr $ \x -> ((0, x), Event x))

If I am not mistaken, Yampa solves this issue with its representation of signal functions. At time zero, the function wrapped in `SF` does not depend on any time step at all, while its continuation wrapped in `SF'` does.

ivanperez-keera:
Yeap, you are absolutely right. I looked at it wrong.

I'm happy accepting this fix. I'll try it with haskanoid first, just in case.

For the next time, can you open the issue first and then mention the issue in the PR comment? It's just to try to keep everything documented and try to follow a clean dev process (i.e., every commit mentions an issue, every commit happens in a branch except a merge commit, commits document the change and motivation clearly, etc.)

ivanperez-keera:
And you are right about Yampa: it enforces the first SF to be zero with the type itself. Unfortunately, as a consequence, Yampa's definitions are always duplicated. So this decision effectively double the size of the codebase.

There must be some smarter way of doing this in Yampa, but that's a separate issue.

ivanperez-keera:
Why the change in `switch`, though? Shouldn't BR's reactimate, if possible, be expressed in terms of BR's language itself (which is the Yampa language)?

ivanperez-keera:
To put it another way: right now, we are using dunai to define BR. That effectively constitutes BR's core.

The smaller that core, the easier it would be to maintain, port, or update BR. The more that we use from dunai, the bigger the conversion layer to connect both.

Linearity:
As it stands `reactimate` doesn't use `BearRiver.switch`, but rather `MSF.switch` plus a mapping between `Maybe` and `Event`. I just tried to make that clearer for myself as I modified `BearRiver.switch`, and I figured it would be illuminating for others too.

I like the sound of minimizing the reliance on Dunai. In this case, I think using `BearRiver.switch` is inappropriate because `senseFirst` and `senseRest` are not naturally signal functions of type `MSF (ReaderT DTime m) a b`.

ivanperez-keera:
Perfect. Slightly amended, but merged. Thanks!

turion:
I'm surprised this works. Why doesn't it divide by zero if we `switch` into `derivative`?

PR #187, opened 2020-01-26
Fix listToMaybeS

### LINKED ISSUE #170: listToMaybeS is incorrect
I believe the function `listToMaybeS` is broken. The correct solution could be:

```
exceptToMaybeS :: Monad m => MSF (ExceptT e m) a b -> MSF (MaybeT m) a b
exceptToMaybeS = morphS $ MaybeT . fmap (either (const Nothing) Just) . runExceptT

listToMaybeS :: Monad m => [a] -> MSF (MaybeT m) arbitrary a
listToMaybeS = exceptToMaybeS . runMSFExcept . sequence . map (step . const . return . (, ()))
```

As a low-hanging fruit for beginners who want to contribute, take this correct code and insert it in `Control.Monad.Trans.MSF.Maybe`, and fix the imports.

turion:
Fixes #170

### DIFF
--- a/src/Control/Monad/Trans/MSF/Maybe.hs
+++ b/src/Control/Monad/Trans/MSF/Maybe.hs
@@ -65,11 +65,15 @@ catchMaybe msf1 msf2 = safely $ do
 
 -- * Converting to and from 'MaybeT'
 
+-- | Convert exceptions into `Nothing`, discarding the exception value.
+exceptToMaybeS :: (Functor m, Monad m) => MSF (ExceptT e m) a b -> MSF (MaybeT m) a b
+exceptToMaybeS = morphS $ MaybeT . fmap (either (const Nothing) Just) . runExceptT
+
 -- | Converts a list to an 'MSF' in 'MaybeT',
 --   which outputs an element of the list at each step,
 --   throwing 'Nothing' when the list ends.
-listToMaybeS :: Monad m => [b] -> MSF (MaybeT m) a b
-listToMaybeS = foldr iPost exit
+listToMaybeS :: (Functor m, Monad m) => [b] -> MSF (MaybeT m) a b
+listToMaybeS = exceptToMaybeS . runMSFExcept . listToMSFExcept
 
 -- * Running 'MaybeT'
 -- | Remove the 'MaybeT' layer by outputting 'Nothing' when the exception occurs.
--- a/dunai.cabal
+++ b/dunai.cabal
@@ -121,3 +121,16 @@ test-suite haddock-coverage
 source-repository head
   type:     git
   location: git@github.com:ivanperez-keera/dunai.git
+
+test-suite regression-tests
+  type: exitcode-stdio-1.0
+  main-is: RegressionTests.hs
+  ghc-options: -Wall
+  hs-source-dirs: tests
+  default-language:  Haskell2010
+  build-depends:
+    base >=4 && <5,
+    transformers,
+    dunai,
+    tasty,
+    tasty-hunit
--- a/src/Control/Monad/Trans/MSF/Except.hs
+++ b/src/Control/Monad/Trans/MSF/Except.hs
@@ -173,6 +173,7 @@ instance Monad m => Applicative (MSFExcept m a b) where
 -- | Monad instance for 'MSFExcept'. Bind uses the exception as the 'return'
 -- value in the monad.
 instance Monad m => Monad (MSFExcept m a b) where
+  return = pure
   MSFExcept msf >>= f = MSFExcept $ handleExceptT msf $ runMSFExcept . f
 
 handleExceptT
--- a/src/Control/Monad/Trans/MSF/Except.hs
+++ b/src/Control/Monad/Trans/MSF/Except.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE Arrows              #-}
 {-# LANGUAGE CPP                 #-}
 {-# LANGUAGE Rank2Types          #-}
+{-# LANGUAGE TupleSections       #-}
 -- | 'MSF's in the 'ExceptT' monad are monadic stream functions
 --   that can throw exceptions,
 --   i.e. return an exception value instead of a continuation.
@@ -225,6 +226,17 @@ step f = try $ proc a -> do
   (b, e) <- arrM (lift . f) -< a
   _      <- throwOn'        -< (n > (1 :: Int), e)
   returnA                   -< b
+
+-- | Advances a single tick outputting the value,
+--   and then throws '()'.
+step_ :: Monad m => b -> MSFExcept m a b ()
+step_ = step . const . return . (, ())
+
+-- | Converts a list to an 'MSFExcept',
+--   which outputs an element of the list at each step,
+--   throwing '()' when the list ends.
+listToMSFExcept :: Monad m => [b] -> MSFExcept m a b ()
+listToMSFExcept = mapM_ step_
 
 -- * Utilities definable in terms of 'MSFExcept'
 
--- a/tests/RegressionTests.hs
+++ b/tests/RegressionTests.hs
@@ -1,0 +1,20 @@
+module Main where
+
+-- transformers
+import Control.Monad.Trans.Class (lift)
+import Control.Monad.Trans.Maybe
+import Control.Monad.Trans.Writer.Lazy
+
+-- dunai
+import Control.Monad.Trans.MSF.Maybe
+import Data.MonadicStreamFunction
+
+-- tasty
+import Test.Tasty
+import Test.Tasty.HUnit
+
+main :: IO ()
+main = defaultMain $ testGroup "listToMaybeS"
+    [ testCase "First emits all list elements before throwing exception" $
+        "Hello" @?= (execWriter $ runMaybeT $ flip embed [(1 :: Integer)..] $ listToMaybeS ["H", "el", "lo"] >>> arrM (tell >>> lift))
+    ]


ivanperez-keera:
Hi! Thanks.

The build broke with:
```
src/Control/Monad/Trans/MSF/Maybe.hs:70:36:
    Could not deduce (Functor m) arising from a use of `fmap'
    from the context (Monad m)
      bound by the type signature for
                 exceptToMaybeS :: Monad m =>
                                   MSF (ExceptT e m) a b -> MSF (MaybeT m) a b
      at src/Control/Monad/Trans/MSF/Maybe.hs:69:19-72
    Possible fix:
      add (Functor m) to the context of
        a type expected by the context: ExceptT e m c -> MaybeT m c
```

Can you add the necessary constraint?

turion:
Done. I always forget that we still build dunai with some GHCs that don't have the Functor-Monad-Hierarchy yet.

ivanperez-keera:
I asked just yesterday about this on fb, twitter and reddit, and there seems to be a number of people who still use 7.*, so I think it's a good idea to keep supporting it.

src/Control/Monad/Trans/MSF/Maybe.hs
```diff
@@ -65,11 +66,14 @@ catchMaybe msf1 msf2 = safely $ do
 
 -- * Converting to and from 'MaybeT'
 
+exceptToMaybeS :: (Functor m, Monad m) => MSF (ExceptT e m) a b -> MSF (MaybeT m) a b
+exceptToMaybeS = morphS $ MaybeT . fmap (either (const Nothing) Just) . runExceptT
```
ivanperez-keera: For info only: what goes in the fmap is:
- http://hackage.haskell.org/package/extra-1.6.18/docs/src/Data.Either.Extra.html#eitherToMaybe
- http://hackage.haskell.org/package/either-4.3.3.2/docs/src/Data-Either-Combinators.html#rightToMaybe
- It also happens to match the behaviour, but not the signature, of https://hackage.haskell.org/package/base-4.12.0.0/docs/src/Data.Maybe.html#listToMaybe. It can also be implemented as `(listToMaybe . toList)`.

It probably makes sense to have something this generic defined somewhere in base, probably in foldable.

I'll happily accept it as is, but it may be useful to know that this expression can be simplified.

src/Control/Monad/Trans/MSF/Maybe.hs
```diff
@@ -65,11 +66,14 @@ catchMaybe msf1 msf2 = safely $ do
...
+
 -- | Converts a list to an 'MSF' in 'MaybeT',
 --   which outputs an element of the list at each step,
 --   throwing 'Nothing' when the list ends.
-listToMaybeS :: Monad m => [b] -> MSF (MaybeT m) a b
-listToMaybeS = foldr iPost exit
+listToMaybeS :: (Functor m, Monad m) => [a] -> MSF (MaybeT m) arbitrary a
+listToMaybeS = exceptToMaybeS . runMSFExcept . sequence . map (step . const . return . (, ()))
```
ivanperez-keera: What is the type of: `exceptToMaybeS . runMSFExcept`?
What is the type of `sequence . map (step . const . return . (, ()))`, and what does it do?

ivanperez-keera: So, when I try to think about this function, it helps me to group this:
```
runMSFExcept . sequence . map (step . const . return . ((\x -> ((), x))))
  :: Monad m => [a1] -> MSF (ExceptT [a1] m) a2 ()
```
What I do not fully understand is why is conceptually correct. This seems to throw an exception with the type of the list at the end.

src/Control/Monad/Trans/MSF/Maybe.hs
turion: I think this doesn't match the behaviour of `listToMaybe`. `listToMaybe` forgets the tail of a list.

src/Control/Monad/Trans/MSF/Maybe.hs
turion: The exception values are discarded. I'll document that.

> What is the type of `sequence . map (step . const . return . (, ()))`, and what does it do?

It outputs each list element and then throws an exception. I could refactor it as a function in `Control.Monad.Trans.MSF.Except` if you want.

turion:
The build failures suggest that exception handling was unusable on GHC <=7.8 all the time and we didn't notice.

src/Control/Monad/Trans/MSF/Maybe.hs
ivanperez-keera: I only mean `either (const Nothing) Just`.

turion: Ah, I see. Yes, I was tempted to use these functions from `extra` or `either`, but didn't want to add a dependency. If you prefer an additional dependency, we can do that as well.

src/Control/Monad/Trans/MSF/Maybe.hs
ivanperez-keera: - `sequence . map` is `mapM`.

- Is sequence . map (step . const . return . (, ())) different from `once . const . return`?

ivanperez-keera: > I could refactor it as a function in `Control.Monad.Trans.MSF.Except` if you want.

No need. I'm just trying to understand it and document it. It may help to give names or types to local definitions in a `where` clause.

src/Control/Monad/Trans/MSF/Maybe.hs
ivanperez-keera: Nah, it's fine. It may be more appropriate to suggest to the libraries people to give this a name and put it in foldable. The definition is small enough that we can just add it, even if there's a shorter way or is defined in some library (just one small definition does not warrant adding a dependency like that).

There is some relation between foldables, traversables and streams that is bugging me and I can't see clearly, but would make this implementation trivial.

src/Control/Monad/Trans/MSF/Maybe.hs
turion: >     * `sequence . map` is `mapM`.

Good point. Fixed. We can even take `mapM_`, to emphasize that the returned list is irrelevant, and only the effect of outputting the elements is important.
 
>     * Is `sequence . map (step . const . return . (, ()))` different from `once . const . return`?

Yes. `once . const . return` does one tick only. But we want to perform one tick for every list element.

src/Control/Monad/Trans/MSF/Maybe.hs
turion: I don't see how a function of type `Either e a -> Maybe a` is related to `Foldable` in a meaningful way.

There is definitely a relation between `Foldable`, `Traversable` and `MSF Maybe () a`. A `Foldable` is simply any list structure (without preferring a particular shape like linked lists or balanced trees over others). A `Traversable` is a `Foldable` that is compatible with `Functor`. I think `MSF Maybe () a` is both, because it's basically a list. Compare the signature of `accumulateWith` with `foldr`.

ivanperez-keera: > I don't see how a function of type `Either e a -> Maybe a` is related to `Foldable` in a meaningful way.

There are ways to implement it in which the type would be more general than that. For one example, the implementation in https://hackage.haskell.org/package/base-4.12.0.0/docs/src/Data.Maybe.html#listToMaybe does not mention `Either` at all.

It is too small a piece of code to argue about it. With my original message, I only wanted to point out that this is very specific and there's probably code somewhere that does this for more general types, or that this could be generally useful for other types. If the solution is not immediately obvious, we don't need to act on this and I don't want to open an issue about it. It'll come up again as we clean the code more and more.

src/Control/Monad/Trans/MSF/Maybe.hs
ivanperez-keera: I have trouble understanding this. I get it, but I don't get it.

I tried to run it on my machine, and the result I got was the same.

ivanperez-keera:
> The build failures suggest that exception handling was unusable on GHC <=7.8 all the time and we didn't notice.

I am not surprised. I never became familiar enough with the implementation of ExceptT in dunai, and I don't really use it, except for the part that bearriver may be needing, if any.

turion:
> > The build failures suggest that exception handling was unusable on GHC <=7.8 all the time and we didn't notice.
> 
> I am not surprised. I never became familiar enough with the implementation of ExceptT in dunai, and I don't really use it, except for the part that bearriver may be needing, if any.

That's funny. I find exception handling one of the prime achievements of Dunai.

src/Control/Monad/Trans/MSF/Maybe.hs
turion: Does it help if I break it up into understandable pieces? This is a big function.

ivanperez-keera: It would, yes. Can you?

It's perfect if the "pieces" are in a where in the same definition. There's no need for now to make them top-level.

ivanperez-keera:
I don't doubt that they are useful. I just could not learn to use them well from the documentation we have right now.

Something that would help me is:
- If the API was grouped based on the types functions operate on (`MSFExcept` vs `MSF ExceptT`).
- If the organization in the file clearly presented a well defined interface (if there is an order in the definitions, it's not immediately obvious to me).
- If the documentation at the top was presented in very simple terms. For example:
```
Generally speaking, MSFs on the Either monad are MSFs that terminate with a
result.

This is an appropriate way of defining MSFs in "steps": first, transformation
one is applied, until it terminates (you are *left* with a result), then we
start applying transformation two until that one terminates (with a result),
and so on. The "results" can be helpful to create the next transformation in
the chain.

It is sometimes convenient to write MSFs in monadic style, such as:
do r1 <- msf1
   r2 <- msf2 r1
   r3 <- msf2 r2
   ... -- and so on

where r1, r2, r3, etc. are the values with which each MSF terminates. To
facilitate using that style, we provide the type MSFExcept, which is
isomorphic to an MSF with an EitherT layer, but with the arguments in the
right order so that they form a monad like the one above. For example:

<show absolutely trivial example here>
```

Points 1 and 2 will happen now. I consider improving the documentation what will come next, but I don't want to spread myself too thin. I want to focus on one thing at a time and forget the documentation for now.

src/Control/Monad/Trans/MSF/Maybe.hs
turion: Turns out I could find some good meaning for the pieces. I made them top level now for haddock, if that's ok.

PR #207, opened 2020-04-21
Fix & CI classicfrp example

turion:


ivanperez-keera:
@turion I think this may work now that #208 is fixed.

Could you please rebase it on top of the new develop and document your change both with an issue that clearly explains the comment and in the commit message?

(Reasoning: Trying to keep everything well documented :) )

turion:
Closes #214.

ivanperez-keera:
The latest change adds dunai-test, but it breaks the CI. It appears to be a missing `Functor` constraint. Can you please check and get the CI to build correctly before I evaluate the PR for merging?

ivanperez-keera:
Rebased and merged into develop.

Issue #208, opened 2020-04-21
Dunai-test is missing Functor constraint


ivanperez-keera:
The dunai-test library is reporting a build error in travis (https://travis-ci.org/github/ivanperez-keera/dunai/jobs/677829694):

```
src/FRP/Dunai/LTLPast.hs:41:5:

    Could not deduce (Functor m) arising from a use of `catchMaybe'
```

To make it compile with older versions of GHC, we could just add this little constraint.

PR #209, opened 2020-04-25
Add Functor Constraint to untilSF

bgluzman:
Hi! This adds the `Functor` constraint recommended in #208 for older versions of GHC.


ivanperez-keera:
Perfect! Thank you, @bgluzman !

PR #210, opened 2020-04-26
Added notice about implementing vector space using linear

walseb:
Since `VectorSpace` in bearriver and in yampa are so similar, the same snippet discussed here https://github.com/ivanperez-keera/Yampa/issues/132 works in bearriver also


ivanperez-keera:
Like in the case of Yampa, the commit has been respected but I reworded the log message so the hash has changed.

Thanks!

Issue #212, opened 2020-04-29
Bearriver's Event is missing the Functor, Monad and Applicative instances


ivanperez-keera:
Bearriver's Event is missing the Functor, Monad and Applicative instances, which exist in Yampa and are very useful.

This is a great candidate for something that could be introduced by a near beginner: the instances are the same as Maybe's, with the following replacement: `Just` => `Event`, `Nothing` => `NoEvent`, and `Maybe` = `Event`.

Just know that, because bearriver and dunai need to be backwards compatible with GHC versions prior to GHC 8.6, you can't use `DerivingGeneric` or `DerivingVia`.

I'm marking this for the next milestone, but it can also be a small release that goes out very soon.

walseb:
Functor and Applicative are [already in](https://github.com/ivanperez-keera/dunai/blob/develop/examples/bearriver/src/FRP/BearRiver.hs#L179), I will try making a monad instance to get to know events better. Thanks for creating these issues!

geroldmeisinger:
Why do we need `Event` instead of `Maybe` anyway if they are "isomorphic"? Seems like there are a lot of other instances missing as well, https://hackage.haskell.org/package/base-4.17.0.0/docs/Prelude.html#t:Maybe

ivanperez-keera:
In bearriver, the reason to have `Event` is because Yampa has them. Bearriver needs to mimic Yampa's interface.

In Yampa,the reason to have `Event` is to semantically indicate with the type what you mean (declarative programming). Also, the `Event` and `EventS` modules in Yampa are very expressive, and you can go a long way without having to use the constructors. So, you can treat `Event` as an abstract data type. If tomorrow we find that the best way to improve Yampa is to change the `Event` constructors, your program doesn't have to change.

If there is an instance that exists in Yampa and not here, file an issue here. If there is an instance that `Maybe` has, `Event` doesn't have, and you need, file an issue in Yampa's repo.

geroldmeisinger:
> In Yampa,the reason to have `Event` is to semantically indicate with the type what you mean (declarative programming).

But couldn't we use an type alias as well?

> If tomorrow we find that the best way to improve Yampa is to change the `Event` constructors, your program doesn't have to change.

..as long as the constructor doesn't change. In the meantime, why not use an alias?

ivanperez-keera:
> > If tomorrow we find that the best way to improve Yampa is to change the Event constructors, your program doesn't have to change.
[...]

PR #213, opened 2020-04-29
Added monad instance, changed functor style

walseb:
https://github.com/ivanperez-keera/dunai/issues/212
So I made use of the `_` wildcard here because the applicative also did. And I made `fmap` infix to match the the other instances more. But I'm not sure if I used the best style here.

### DIFF
--- a/examples/bearriver/src/FRP/BearRiver.hs
+++ b/examples/bearriver/src/FRP/BearRiver.hs
@@ -175,16 +175,19 @@ afterEachCat = afterEachCat' 0
 
 
 -- * Events
-
 instance Functor Event where
-  fmap f NoEvent   = NoEvent
-  fmap f (Event c) = Event (f c)
+   f `fmap` (Event c) = Event (f c)
+   _ `fmap` _         = NoEvent
 
 instance Applicative Event where
   pure = Event
-
   Event f <*> Event x = Event (f x)
   _       <*> _       = NoEvent
+
+instance Monad Event where
+  return          = pure
+  (Event x) >>= f = f x
+  _         >>= _ = NoEvent
 
 -- | Apply an 'MSF' to every input. Freezes temporarily if the input is
 -- 'NoEvent', and continues as soon as an 'Event' is received.
--- a/examples/bearriver/src/FRP/BearRiver.hs
+++ b/examples/bearriver/src/FRP/BearRiver.hs
@@ -176,18 +176,19 @@ afterEachCat = afterEachCat' 0
 
 -- * Events
 instance Functor Event where
-   f `fmap` (Event c) = Event (f c)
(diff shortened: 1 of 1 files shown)


examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -175,17 +175,20 @@ afterEachCat = afterEachCat' 0
...
 
 instance Applicative Event where
   pure = Event
-
   Event f <*> Event x = Event (f x)
   _       <*> _       = NoEvent
 
+instance Monad Event where
+  return          = pure
+  (Event x) >>= f = f x
```
turion, line 189: You can drop the parentheses around `Event x`

examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -175,17 +175,20 @@ afterEachCat = afterEachCat' 0
 
 
 -- * Events
-
 instance Functor Event where
-  fmap f NoEvent   = NoEvent
-  fmap f (Event c) = Event (f c)
+   f `fmap` (Event c) = Event (f c)
```
turion, line 179: I personally find this a bit awkward to read and don't see the advantage.

examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -175,17 +175,20 @@ afterEachCat = afterEachCat' 0
...
+   _ `fmap` _         = NoEvent
 
 instance Applicative Event where
   pure = Event
-
   Event f <*> Event x = Event (f x)
   _       <*> _       = NoEvent
 
+instance Monad Event where
```
turion, line 187: If you want you can add a haddock in the line before that says that this instance is the same as `Maybe`'s.

examples/bearriver/src/FRP/BearRiver.hs
```diff
@@ -175,17 +175,20 @@ afterEachCat = afterEachCat' 0
...
+  return          = pure
+  (Event x) >>= f = f x
+  _         >>= _ = NoEvent
```
turion, line 190: I would find it easier to reason about if you mention the pattern here.

walseb:
Thanks! Everything should be fixed now

ivanperez-keera:
Hi, @walseb . Thanks a lot!!

I have merged these commits into the develop branch to be included in the next release.

I made some adaptations, both to the commit message, and to the style. I never really formalized any of those, so they "feel right" and I could argue why I prefer them, but it's more of a soft preference than a hard rule that I'm applying. If you want, take a look :)

As always, commit authorship of the commits from you that I modified (squashed & message reworded) was maintained.

ivanperez-keera:
Thank you also @turion for the comments and contribution :)

PR #215, opened 2020-05-01
Cabal project

### LINKED ISSUE #205: Simplify Travis setup with cabal projects
In rhine, I recently vastly reduced the number of Travis CI jobs - thus speeding up the pipeline - by building all packages in one job using cabal projects. Here is an example: https://github.com/turion/rhine/blob/9dbcf9cba810b88a4685f473ceaf0af77c1c5bc3/.travis.yml

This may or may not be applicable in Dunai as well. Some packages failed in specific GHC versions, and I don't know whether they still do.

turion:
Supersedes #166. Fixes #205 and #165.

This will probably not work because some packages will fail to build on older GHC versions. My personal opinion is to drop support for these versions, but I understand that you, @ivanperez-keera, want to keep the support. Another possibility might be to fix these packages for those older GHC versions. Might require CPP though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ env:
   - PACKAGEDIR="."
   - PACKAGEDIR="examples/bearriver"
   - PACKAGEDIR="examples/bouncingball"
+  - PACKAGEDIR="examples/classicfrp"
   - PACKAGEDIR="examples/discrete"
   - PACKAGEDIR="examples/keerahails"
   - PACKAGEDIR="examples/list"
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ ghc:
 env:
   - PACKAGEDIR="."
   - PACKAGEDIR="examples/bearriver"
+  - PACKAGEDIR="extensions/dunai-test"
   - PACKAGEDIR="examples/bouncingball"
   - PACKAGEDIR="examples/classicfrp"
   - PACKAGEDIR="examples/discrete"
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,47 +13,13 @@ ghc:
   - "8.6.5"
   - "8.8.2"
 
-env:
-  - PACKAGEDIR="."
-  - PACKAGEDIR="examples/bearriver"
-  - PACKAGEDIR="extensions/dunai-test"
-  - PACKAGEDIR="examples/bouncingball"
-  - PACKAGEDIR="examples/classicfrp"
-  - PACKAGEDIR="examples/discrete"
-  - PACKAGEDIR="examples/keerahails"
-  - PACKAGEDIR="examples/list"
-  - PACKAGEDIR="examples/taggingmonad"
-  - PACKAGEDIR="examples/reversefrpzoo/wormholes"
-
-matrix:
-  exclude:
-    - ghc: 7.6.3
-      env: PACKAGEDIR="examples/list"
-    - ghc: 7.8.4
-      env: PACKAGEDIR="examples/discrete"
-    - ghc: 7.8.4
-      env: PACKAGEDIR="examples/keerahails"
-    - ghc: 8.2.2
-      env: PACKAGEDIR="examples/discrete"
-    - ghc: 8.2.2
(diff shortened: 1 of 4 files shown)


ivanperez-keera:
I still do not understand the purpose of a `cabal.project` file and how to use it. Dunai does not have any dependencies internally in this directory, so I don't get why a `cabal.project` file would be really needed or even appropriate.

(I'm not opposed, I'm just ignorant of the benefits and the intended usage.)

Separately from that, the build is currently broken for several reasons. On older GHC versions, the first break occurs with the missing `Functor` constraint. On newer GHC versions, it happens at the first mention of `wx`. Some would need to be addressed, others, omitted.

turion:
> I still do not understand the purpose of a cabal.project file and how to use it.

Usage: Run `cabal build all && cabal test all` in the root directory.

Purposes:

* Build all projects at once.
* Wildcards in the project file mean we're less likely to accidentally omit subpackages from CI.
* Reducing rebuild times of internal dependencies.
* Fewer CI jobs.

> Dunai does not have any dependencies internally in this repository,

But isn't `dunai` itself an internal dependency? And `bearriver` likewise?

Each CI job needs to

1. Boot the VM
2. Install external dependencies (GHC, libraries, hackage dependencies)
3. Build `dunai`
4. Build the specific package

By using `cabal.project`, we save on steps 1. - 3. and have to run through them only once per GHC version.

> Separately from that, the build is currently broken for several reasons. On older GHC versions, the first break occurs with the missing Functor constraint. On newer GHC versions, it happens at the first mention of wx. Some would need to be addressed, others, omitted.

By omitting, do you mean removing from CI? Or maybe putting them in a separate repository, e.g. `dunai-examples`?

turion:
Hopefully I've fixed the Functor constraint by rebasing onto #207 .

ivanperez-keera:
Not there yet. There's some issues with src/FRP/Dunai/QuickCheck.hs (due to not importing applicative) and one of the examples uses wxcore, which fails to build.

ivanperez-keera:
This is more of an open discussion about cabal itself than addressing the issue at hand. I'm not criticizing your effort, I'm just a bit puzzled by the direction cabal has taken with this. So, apologies in advance if it seems like I'm diminishing your PR: I am not.

I see how listing them in a file may be useful, I just don't really understand cabal's move towards "project" files and the perception that if you don't have a project file you don't support the "new way of building projects" (see https://github.com/ivanperez-keera/dunai/issues/165).

Also, it's as easy as:
```bash
$ cabal sandbox init
$ find . -iname '*.cabal' -exec dirname {} \; | xargs cabal sandbox add-source 
$ find . -iname '*.cabal' -exec dirname {} \; | cabal install --enable-tests --run-tests
```

No? I think you can actually simplify it as:
```bash
$ export ALL_PACKAGES=$(find . -iname '*.cabal' -exec dirname {} \;)
$ cabal sandbox init
$ cabal sandbox add-source $ALL_PACKAGES
$ cabal install --enable-tests --run-tests $ALL_PACKAGES
```

I'm confident in some new cabal version you can do something along the lines of:
```bash
$ export ALL_PACKAGES=$(find . -iname '*.cabal' -exec dirname {} \;)
$ cabal v2-install --enable-tests --run-tests $ALL_PACKAGES
```

turion:
For the `wxcore` issue, it's your decision whether to take those packages from the pipeline or to put them in a separate repository. (Or to fix `wxcore` somehow...)

I can chase down all the Functor things eventually. It just needs a few iterations.

I have no special feelings towards `cabal.project`. I'm only after a way to shorten CI time. If your approach does that as well, then maybe you should add that to the Travis script.

ivanperez-keera:
I think that example can be disabled (that's precisely why it is disabled in the current build).

ivanperez-keera:
Thanks! I have worked on this and merged a version of this file.

For future reference, this is what I had to do and the kind of thing that would speed up merges and PR approval:

- Getting the CI to pass.
- Adding a more detailed description to the commit.
- Documenting the changes separately, and adding different PRs for each.

Issue #223, opened 2020-05-10
Adding the bearriver arcade to readme


walseb:
Hi! I just published a bunch of arcade(ish) games I made made using bearriver! It shows you of the whole process from input management, game saving and loading, keeping a target framerate to rendering stuff with rotation and scale using SDL. Maybe bearriver should get a readme similar to yampa displaying these kinds of projects. Or maybe they should all be kept in the dunai readme since it's based on it. This is probably related to #220 
https://github.com/walseb/The_Bearriver_Arcade

One thing to note is that the games doesn't actually use anything that's not in yampa, but one of the games I'm working on does

ivanperez-keera:
This is wonderful!!

Bearriver does not have a README (yet), but I agree that they should be added to Dunai's README. It could make sense to also add it to Yampa's: they could be useful for both projects!

walseb:
Just a heads up for anyone trying it out, I just fixed a bug where the window didn't resize properly causing it to be unplayable on any machine but mine (because my window manager forces all windows into full-screen mode)

ivanperez-keera:
@walseb Feel free to send a PR. I'd be happy to add this. These are fun to play, and useful resources for anyone who wishes to build their own.

PR #226, opened 2020-05-16
Added notice to readme about the bearriver arcade

walseb:
#223

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,9 @@ The following papers are also related to MSFs:
 
 - [Rhine: FRP with Type-level Clocks](https://www.youtube.com/watch?v=Xvgz11D7xqs). Talk describing how MSFs can be extended with clocks. Haskell Symposium 2018.
 
+## Games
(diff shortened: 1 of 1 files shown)


ivanperez-keera:
Thanks!

To keep things documented, could I ask you that you please refer, in the commit message, to the issue that you opened, with the text `Refs #223` ?

Also, I'm not sure what more can be said that could be meaningful, but if you can think of something else, you can add it to the same message. Just imagine the github repo was lost and someone that's never seen dunai had to go through the commit messages to understand what happened. What would you tell them that could be helpful beyond the top line? If anything, then please write that too :)

walseb:
Sure! I will fix it tomorrow

walseb:
I have changed the commit messages in this repo and yampa now

ivanperez-keera:
Hi! Thanks!

I have modified it a bit (both the message and the entry), but your authorship is respected :) It's merged but github won't detect it because I amended the commit message.

PR #231, opened 2020-06-20
FIxed dlSwitch typo in bearriver list module

### LINKED ISSUE #230: bearriver: Broken implementation of dlSwitch
There appears to be a typo in the bearriver list module since `dlSwitch` is missing an implementation while `dlSwitch'` is defined just below without an explicit type signature. I'm not sure which is using the correct name because the function is not present in yampa.

Location:
https://github.com/ivanperez-keera/dunai/blob/develop/dunai-frp-bearriver/src/FRP/BearRiver/List.hs#L17

walseb:
Closes #230

### DIFF
--- a/dunai-frp-bearriver/src/FRP/BearRiver/List.hs
+++ b/dunai-frp-bearriver/src/FRP/BearRiver/List.hs
@@ -15,7 +15,7 @@ import           FRP.BearRiver
 newtype ListSF a b = ListSF { listSF :: SF a (b, Bool, [ListSF a b]) }
 
 dlSwitch :: [ListSF a b] -> SF a [b]
-dlSwitch' sfs = MStreamF $ \a -> do
+dlSwitch sfs = MStreamF $ \a -> do
   -- results of applying the initial input
   bsfs0 <- mapM ((`unMStreamF` a).listSF) sfs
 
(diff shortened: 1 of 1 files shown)


dunai-frp-bearriver/src/FRP/BearRiver/List.hs
```diff
@@ -15,7 +15,7 @@ import           FRP.BearRiver
 newtype ListSF a b = ListSF { listSF :: SF a (b, Bool, [ListSF a b]) }
 
 dlSwitch :: [ListSF a b] -> SF a [b]
```
turion, line 17: How could this have compiled before without a value bound to this name?

walseb: I don't think it is compiled, it's not mentioned in the cabal file or any imports I can find. Or maybe missing implementation is just a warning and not an error?

ivanperez-keera:
There were other errors in this implementation. It's now been fixed and included.

The commit log explains why it is not compiled.

See also https://github.com/ivanperez-keera/Yampa/blob/develop-games/src/FRP/Yampa/Switches.hs#L1008-L1052.

ivanperez-keera:
Thanks @walseb !

ivanperez-keera:
@walseb I tried to make it so that git would also credit the commit to you (see the commit log), but apparently it did not. Sorry about that. It was not on purpose.

walseb:
@ivanperez-keera No problem!

PR #240, opened 2020-12-17
Move export of `VectorSpace` to new module

Linearity:
To date, `FRP.BearRiver` re-exports the module `Data.VectorSpace` from the "simple-affine-space" package. That makes it easy to use its definitions like `^+^` just by importing `FRP.BearRiver`. However, these names crowd the namespace, and if another vector package uses the same names we have to qualify them. For example, one cannot import both `FRP.BearRiver` and `Linear` (from the "linear" package) without qualifying names.

In this branch, `FRP.BearRiver` does not export `Data.VectorSpace`; this is done by a new module, `FRP.BearRiver.VectorSpace`. The main module still permits the use of `integral` on all instances of `VectorSpace` defined there, and we don't have to worry about conflicting definitions of `^+^`. Any code that uses the `Data.VectorSpace.^+^` or defines new instances must additionally import `FRP.BearRiver.VectorSpace`.

This makes it much less hassle to use Bear River with other vector packages in the general case, at the cost of requiring an extra import statement in some cases.


dunai-frp-bearriver/src/FRP/BearRiver/VectorSpace.hs
```diff
@@ -0,0 +1,3 @@
+module FRP.BearRiver.VectorSpace (module Data.VectorSpace) where
+
+import Data.VectorSpace
```
turion, line 3: Missing newline

turion (changes requested):

dunai-frp-bearriver/src/FRP/BearRiver/VectorSpace.hs
Linearity: Done.

dunai/stack.yaml
```diff
@@ -1 +1,3 @@
 resolver: lts-11.7
+
+extra-deps: [simple-affine-space-0.1]
```
dpwiz, line 3: btw, simple-affine-space is on stackage now

turion: Yes, then it should be removed from the stack file.

ivanperez-keera:
If there's any change in this respect in Bearriver, it should be incorporated in Yampa first.

Also, why not just:

```haskell
import FRP.BearRiver hiding (VectorSpace(..))
```

The following works for me:

```haskell
import FRP.BearRiver hiding (VectorSpace(..))
import Linear

test :: Monad m => SF m (V2 Double, V2 Double) (V2 Double)
test = arr (uncurry (^+^))

same :: (Additive v, Num n, Monad m) => SF m (v n) (v n)
same = arr (^+^ zero)
```

You can always just drop this file in your projects:

```haskell
module FRP.BearRiver.Linear
    ( module BR
    , module L
    )
  where

import FRP.BearRiver as BR hiding (VectorSpace(..), identity)
import Linear        as L
```

which you never have to touch, and then you can just:

```haskell
import FRP.BearRiver.Linear

test :: Monad m => SF m (V2 Double, V2 Double) (V2 Double)
test = arr (uncurry (^+^))

same :: (Additive v, Num n, Monad m) => SF m (v n) (v n)
same = arr (^+^ zero)
```

I'm closing this issue in favor of either the solution proposed or, if the solution is not good enough, opening an issue in Yampa documenting the problem. Please open the issue first to discuss. If you send a PR without an issue that documents the problem and where we agree on it being included, it'll unfortunately have to be rejected.

Thanks!

PR #241, opened 2020-12-17
Support for signal functions in transformed monads 

### LINKED ISSUE #174: BearRiver's `SF` does not use `MonadReader`
BearRiver represents signal functions with a type that is polymorphic in the monad defining the type of the underlying monadic stream function, namely

    type SF m = MSF (ReaderT DTime m)

This is nice because it offers users the opportunity to incorporate other monadic effects. But like many applications of monad transformers it forces one particular transformer, `ReaderT DTime`, to be at the top of whatever transformer stack you might want to use. That means that one must use `lift` to incorporate, say, state handling in signal functions.

If it instead used a `MonadReader DTime m` constraint in the style of the "mtl" package, it could be more flexible. [There is a demand for this flexibility.](http://blog.ezyang.com/2013/09/if-youre-using-lift-youre-doing-it-wrong-probably/) Is this a reasonable thing to do in BearRiver?

The actual type synonym `SF` could not be defined this way, but the primitive signal functions could be. For instance, the type of `edge` could be:

    edge :: MonadReader DTime m => MSF m Bool (Event ())

Then it would be easy to use `edge` in a signal function whose monad has `StateT` at the top. This would be nice, because then we could apply `runStateS__` before passing the result to `reactimate`, which naturally applies `runReaderS`. As it stands, I believe, we must apply `runStateS__` _after_ `reactimate`, meaning that the entirety of the composite signal function could use only one state type for everything. That is a tremendous limitation, and it seems relatively straightforward to remove.

linearity:
`SF` has a monad parameter that is transformed with `ClockInfo` in its type synonym definition:

> type SF m = MSF (ClockInfo m)

This branch defines transformations of signal functions between some different monad parameters. These make it easier to transform this monad "underneath" the `ClockInfo` layer. It is inspired by [Rhine's transformer-commuting code][rhine].

Additionally, it changes `reactimate` to work on signal functions with monads other than `Identity`.

Fixes #174

[rhine]: http://hackage.haskell.org/package/rhine-0.5.1.1/docs/FRP-Rhine-ClSF-Random.html


dunai-frp-bearriver/src/FRP/BearRiver.hs
```diff
@@ -617,3 +617,109 @@ replaceOnce a = dSwitch (arr $ const (a, Event ())) (const $ arr id)
 
 -- ** Tuples
 dup  x     = (x,x)
+
+-- * Monad transformations
```
turion, line 621: Can you put this section in a separate file? @ivanperez-keera maybe has a good idea for a name.

dunai-frp-bearriver/src/FRP/BearRiver.hs
```diff
@@ -617,3 +617,109 @@ replaceOnce a = dSwitch (arr $ const (a, Event ())) (const $ arr id)
...
+
+-- | Lift an 'SF' in a certain monad to one in the same monad transformed
+-- with 'WriterT'
+liftWriterSF :: (Monad m, Monoid w) => SF m a b -> SF (WriterT w m) a b
+liftWriterSF sf = morphS commuteReaderWriter (liftTransS sf)
+
+-- | Transform an action in a monad transformer stack with a 'WriterT' layer
+-- on top of a 'ReaderT' layer into one in another stack where the layers are
+-- swapped
+commuteReaderWriter (WriterT m)
```
turion, line 718: Add a type signature.
Also, this function should not be exported from the module.

dunai-frp-bearriver/src/FRP/BearRiver.hs
```diff
@@ -617,3 +617,109 @@ replaceOnce a = dSwitch (arr $ const (a, Event ())) (const $ arr id)
...
+
+-- | Build an 'SF' /function/ that takes a fixed environment as additional
+-- input, from an SF in the 'Reader' monad
+runReaderSF_ :: Monad m => r -> SF (ReaderT r m) a b -> SF m a b
+runReaderSF_ r sf = runReaderS_ (morphS commuteReaderReader sf) r
+
+-- | Lift an 'SF' in a certain monad to one in the same monad transformed
+-- with 'ReaderT'
+liftReaderSF :: Monad m => SF m a b -> SF (ReaderT r m) a b
+liftReaderSF sf = morphS commuteReaderReader (liftTransS sf)
```
turion, line 647: See https://github.com/turion/rhine/blob/686632679915c0415a619ff7ff5ab557fc9c08f4/rhine/src/FRP/Rhine/ClSF/Core.hs#L80 and https://github.com/turion/rhine/blob/686632679915c0415a619ff7ff5ab557fc9c08f4/rhine/src/FRP/Rhine/ClSF/Core.hs#L68. You shouldn't have to implement a lifting function for every transformer.

turion (changes requested):

dunai-frp-bearriver/src/FRP/BearRiver.hs
Linearity: Done. I just called it `FRP.BearRiver.Monad`.

dunai-frp-bearriver/src/FRP/BearRiver.hs
Linearity: Done. None of the commuting functions are exported.

dunai-frp-bearriver/src/FRP/BearRiver.hs
Linearity: I just based this on `FRP.Rhine.ClSF.Random` and wasn't aware of a different technique. Can you describe what you're hoping to see?

turion: I expect there is a function `liftSF :: (Monad m, MonadTrans t) => SF m a b -> SF (t m) a b` which should be implemented just like https://github.com/turion/rhine/blob/686632679915c0415a619ff7ff5ab557fc9c08f4/rhine/src/FRP/Rhine/ClSF/Core.hs#L68. This function can replace `liftReaderSF`, `liftStateSF`, and the other specialised lifting functions.

Linearity: Oh you're right. I can use `liftTransSF` anywhere I use `liftReaderSF`, `liftStateSF` or `liftWriterSF`. I removed those and re-implemented `liftTransSF` as `liftSF` using a new function `hoistSF`.

PR #242, opened 2020-12-17
Add primitives `link` and `vain`

Linearity:
This branch defines a `link`, a new switching primitive for Bear River, and `vain`, a continuation-capturing primitive for Dunai.

These can be used to implement high-level switching facilities as described in [this Haskell Symposium paper][paper]. They serve the same purpose as `Task`s with greater flexibility.

In the future I would recommend changing the official `Task` implementation to use these primitives. This branch, however, simply makes them available to other implementations.

[paper]: https://das.li/scripted-signal-functions.pdf

### DIFF
--- a/dunai-frp-bearriver/src/FRP/BearRiver.hs
+++ b/dunai-frp-bearriver/src/FRP/BearRiver.hs
@@ -18,16 +18,17 @@ import qualified Control.Category                               as Category
 import           Control.Monad                                  (mapM)
 import           Control.Monad.Random
 import           Control.Monad.Trans.Maybe
-import           Control.Monad.Trans.MSF                        hiding (switch)
+import           Control.Monad.Trans.MSF                        hiding (link, switch)
 import qualified Control.Monad.Trans.MSF                        as MSF
 import           Control.Monad.Trans.MSF.Except                 as MSF hiding
-                                                                        (switch)
+                                                                        (link, switch)
 import           Control.Monad.Trans.MSF.List                   (sequenceS,
                                                                  widthFirst)
 import           Control.Monad.Trans.MSF.Random
 import           Data.Functor.Identity
 import           Data.Maybe
-import           Data.MonadicStreamFunction                     as X hiding (reactimate,
+import           Data.MonadicStreamFunction                     as X hiding (link,
+                                                                      reactimate,
                                                                       repeatedly,
                                                                       sum,
                                                                       switch,
@@ -436,6 +437,12 @@ dSwitch sf sfC = MSF $ \a -> do
                        return (b, ct')
     (b, NoEvent) -> return (b, dSwitch ct sfC)
 
+link :: Monad m => SF m a (b, Event c) -> (c -> SF m a (b, Event d)) -> SF m a (b, Event d)
+link sf sfC = MSF $ \a -> do
+  (o, ct) <- unMSF sf a
+  case o of
+    (_, Event c) -> local (const 0) (unMSF (sfC c) a)
+    (b, NoEvent) -> return ((b, NoEvent), link ct sfC)
 
 -- * Parallel composition and switching
 
--- a/dunai-frp-bearriver/src/FRP/BearRiver.hs
+++ b/dunai-frp-bearriver/src/FRP/BearRiver.hs
@@ -422,14 +422,24 @@ e `gate` True  = e
 
 -- ** Basic switchers
 
-switch :: Monad m => SF m a (b, Event c) -> (c -> SF m a b) -> SF m a b
+-- | A signal function in two operating modes, one before some event and one
+-- after. During the switching time step the output is that of the second mode.
+switch :: Monad m =>
+            SF m a (b, Event c)     -- ^ 'SF' that outputs switching event stream
+              -> (c -> SF m a b)    -- ^ function that gives the second mode
+              -> SF m a b           -- ^ the combined 'SF'
 switch sf sfC = MSF $ \a -> do
   (o, ct) <- unMSF sf a
   case o of
     (_, Event c) -> local (const 0) (unMSF (sfC c) a)
     (b, NoEvent) -> return (b, switch ct sfC)
 
-dSwitch ::  Monad m => SF m a (b, Event c) -> (c -> SF m a b) -> SF m a b
+-- | A variant of 'switch' where the output during the switching time step is
+-- that of the first mode, not the second mode.
+dSwitch ::  Monad m =>
+              SF m a (b, Event c)   -- ^ 'SF' that outputs switching event stream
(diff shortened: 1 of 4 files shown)


dunai/src/Data/MonadicStreamFunction/InternalCore.hs
```diff
@@ -133,3 +133,8 @@ reactimate :: Monad m => MSF m () () -> m ()
 reactimate sf = do
   (_, sf') <- unMSF sf ()
   reactimate sf'
+
+vain :: Monad m => MSF m a b -> MSF m a (b, MSF m a b)
+vain msf0 = MSF { unMSF = f }
```
turion, line 138: Can you format this in the same way you formatted `link`?

dunai-frp-bearriver/src/FRP/BearRiver.hs
```diff
@@ -436,6 +437,12 @@ dSwitch sf sfC = MSF $ \a -> do
                        return (b, ct')
     (b, NoEvent) -> return (b, dSwitch ct sfC)
 
+link :: Monad m => SF m a (Either b c) -> (c -> SF m a (Either b d)) -> SF m a (Either b d)
+link sf sfC = MSF $ \a -> do
```
turion, line 441: It's always a maintenance effort to add new primitives that use the `MSF` constructor. Have you tried to implement this using exceptions (https://hackage.haskell.org/package/dunai-0.5.2.1/docs/Control-Monad-Trans-MSF-Except.html)?

dunai/src/Data/MonadicStreamFunction/InternalCore.hs
turion, line 138: Similar to my comment for `link`, it's maintenance effort to add new functions that use the `MSF` constructor. In this case, it seems pretty fundamental. Can you express any other functions in term of it?

dunai/src/Data/MonadicStreamFunction/InternalCore.hs
```diff
@@ -133,3 +133,8 @@ reactimate :: Monad m => MSF m () () -> m ()
 reactimate sf = do
   (_, sf') <- unMSF sf ()
   reactimate sf'
+
+vain :: Monad m => MSF m a b -> MSF m a (b, MSF m a b)
```
turion, line 137: This should have a haddock.

dunai-frp-bearriver/src/FRP/BearRiver.hs
```diff
@@ -436,6 +437,12 @@ dSwitch sf sfC = MSF $ \a -> do
                        return (b, ct')
     (b, NoEvent) -> return (b, dSwitch ct sfC)
 
+link :: Monad m => SF m a (Either b c) -> (c -> SF m a (Either b d)) -> SF m a (Either b d)
```
turion, line 440: This should have a haddock.

turion (changes requested):
Can you add usage examples for these functions somewhere that illustrate how they are useful? An example maybe, or a test case?

dunai-frp-bearriver/src/FRP/BearRiver.hs
Linearity: Done. I added them to `switch` and `dSwitch` as well.

dunai/src/Data/MonadicStreamFunction/InternalCore.hs
Linearity: Done.

dunai/src/Data/MonadicStreamFunction/InternalCore.hs
Linearity: Done.

dunai-frp-bearriver/src/FRP/BearRiver.hs
Linearity: I have: you can see my experiment in [one of the versions in this branch][except]. It uses the output type `(b, Event c)` instead of `Either b c`, but that's probably easy to change.

However, it's not obvious to me how to ensure that the time step is zero for the actions after the switching event. This version is simpler and closely matches the implementations of `switch` and `dSwitch`.

[except]: https://github.com/ivanperez-keera/dunai/pull/242/commits/07623ed1e0ed4ab2374cdeff1297e1365a38eac2#diff-9c2aba48d4d293c9782080f4347a9074e906e6476cece93c907958e8282c1a06R276

dunai/src/Data/MonadicStreamFunction/InternalCore.hs
Linearity: You can express parallel switching in terms of `vain`, as in section 4 of the paper.

turion: Great, that sounds useful :) Then such combinators should be added.

Linearity: Can that happen separately, or does that need to be in this branch?

turion: I personally think it should happen here.

turion:
@ivanperez-keera I think we need your opinion to proceed.

ivanperez-keera:
Sorry fro the lag, I spent Xmas completely disconnected from everything and everybody but I'm back now. I'll take a look as soon as I can.

dunai/src/Data/MonadicStreamFunction/InternalCore.hs
Linearity: In that case I have another branch that implements Tasks with `link` and `vain`, and I could merge it into this branch.

dpwiz:
Any plans to unstuck this? (:

walseb:
As a sidenote, does anyone know where the other primitives mentioned in the paper (`mix` for example) are defined? They are implemented in dunai so one would imagine there is a package out there

turion:
Personally I think that a good working example (see https://github.com/ivanperez-keera/dunai/tree/develop/dunai-examples) should be added to illustrate a use case in which existing dunai primitives aren't sufficient.

ivanperez-keera:
I remember reading this and seeing it was very similar to some presentation system we had at Keera built on top of Dunai, but I forgot the details.

I have a simple question for @Linearity : Would it make sense to add this to Yampa? Bearriver in principle tries to mirror (and generalize) the Yampa API, so a natural path for any changes to Bearriver is to add them to Yampa first, and them port them here.

Linearity:
Yes, these things make sense in Yampa. I can make a Yampa branch for them.

ivanperez-keera:
Based on the discussion regarding opening an issue in the Yampa project, I'm closing this PR here for now.

If this feature is added to Yampa, then it can be "flowed down" to bearriver after that.

PR #244, opened 2021-02-05
Stop repeatedly from getting stuck in infinite loop. Close #243.

### LINKED ISSUE #243: Bearriver: afterEachCat doesn't behave like in Yampa
Currently `repeatedly` is defined:

``` haskell
repeatedly :: Monad m => Time -> b -> SF m a (Event b)
repeatedly q x
    | q > 0     = afterEach qxs
    | otherwise = error "bearriver: repeatedly: Non-positive period."
  where
    qxs = (q,x):qxs
```

Given `q = 1` and `x = "foo"` it feeds `afterEach` with an infinite stream of: `[(1, "foo"), (1, "foo"), (1, "foo"), (1, "foo"), ...]` because `qxs = (q,x):qxs` infinitely recurses with the same time and same event

This makes sense if the time fed to `afterEach` is relative to the last event fired. For example if `afterEach` was given `(1, "foo"), (100, "bar")`, an event would fire with `foo` at time 1 and then it schedule the next event to be fired with result `bar` at time 101. But this isn't the case, what actually happens is that `foo` fires at time 1 and `bar` fires at time 100

The problem with this is that if you have an infinite stream of `[(time, ()), (time, ()), ...]` (as the function qxs always generates), they will all fire at whatever `time` is the current time, resulting in an infinite loop. So if `time` is 10, every event in the infinite stream will fire after 10 seconds

The solutiton is to create an infinite lists where the time of the next element is `time * N`: `[(time * 1, ()), (time * 2, ()), (time * 3, ()), ...]` so that you get evenly spaced out events by time `time`. For example:
```
[(1, "foo"), (2, "foo"), (3, "foo"), (4, "foo"), ...]
```

walseb:
#243

I chose to modify repeatedly to fix this, but we could also modify `afterEach` to make it reset its stored time every time an event fires. But I think this would be very unpredictable especially if multiple events fire at the same time, then the timing would be changed for all other future times unless we accumulated the time of the events fired.

### DIFF
--- a/dunai-frp-bearriver/src/FRP/BearRiver.hs
+++ b/dunai-frp-bearriver/src/FRP/BearRiver.hs
@@ -174,7 +174,7 @@ repeatedly q x
     | q > 0     = afterEach qxs
     | otherwise = error "bearriver: repeatedly: Non-positive period."
   where
-    qxs = (q,x):qxs
+    qxs = (\q' -> (q', x)) <$> scanl (+) q (repeat q)
 
 -- | Event source with consecutive occurrences at the given intervals.
 -- Should more than one event be scheduled to occur in any sampling interval,


turion:
Looking good to me.

walseb:
One design decision to make is if this should fire right after switching or if it should wait for time `q` then fire for the first time. Currently it fires for the first time after one time `q` has elapsed

walseb:
I think it's better to not fire at time 0 because it's much easier to combine a `SF` firing an event at time 0 than it is to try to remove the initial event signal from `repeatedly`

turion:
Thanks!

ivanperez-keera:
There's a reason why this function is implemented this way:

https://github.com/ivanperez-keera/Yampa/blob/4aa42764ea1794f7d94967a5a7e1e769c5fdd67d/src/FRP/Yampa/EventS.hs#L156-L166

This function is literally a copy of Yampa's. If the bug exists in Yampa, then the right solution would be to first fix it there, **and then** bring it here.

Bearriver mirrors Yampa, not the other way around. Changes should only travel in one direction (from Yampa to Bearriver, until such time when Bearriver implements all of Yampa as efficiently, and can simply be used instead).

ivanperez-keera:
@walseb Please see if the error also occurs in Yampa. If it does, open an issue and we'll fix it there first. I'll walk you through the changes. We can later bring it to bearriver.

Issue #224, opened 2020-05-12
MSFs are profunctors


turion:
I think that for any functor `m`, `MSF m` is a [profunctor](https://hackage.haskell.org/package/profunctors-5.5.2/docs/Data-Profunctor.html#t:Profunctor). This means that we can pre- and postcompose `MSF`s with pure functions, and that precomposition is contravariant. In arrows, precomposition is known as the [`(^>>)`](https://hoogle.haskell.org/?hoogle=%5E%3E%3E) operator.

For the additional dependency of the `profunctor` package, we could implement a `Profunctor` instance for `MSF m`. Possibly, this is not so hard, and might be a beginner's issue. I'm not sure what the huge advantages are, though.

turion:
Similarly, I believe that `MSF m` is a [`Traversing`](https://hackage.haskell.org/package/profunctors-5.6.2/docs/Data-Profunctor-Traversing.html#t:Traversing): From an `MSF m a b` you can make an `MSF m [a] [b]`, and similarly for other `Traversable` structures.

PR #247, opened 2021-08-21
Add instances for Profunctor and Traversing.

freckletonj:
This is my first pass at adding Profunctor and Traversing instances.

I'm not 100% positive they're correct, though, some hand-rolled experiments seem to match my intuitions so far.

Related: https://github.com/ivanperez-keera/dunai/issues/224


dunai/dunai.cabal
```diff
@@ -82,7 +83,8 @@ library
                      transformers,
                      transformers-base,
                      MonadRandom,
-                     simple-affine-space
+                     simple-affine-space,
+                     profunctors
```
freckletonj, line 87: this would require a new dependency

dunai/src/Data/MonadicStreamFunction/Instances/Profunctor.hs
```diff
@@ -0,0 +1,31 @@
...
+
+instance Monad m => Strong (MSF m) where
+  first' = first
+
+instance Monad m => Profunctor (MSF m) where
+  dimap l r p = arr l >>> p >>> arr r
+
+instance (Monad m) => Traversing (MSF m) where
+  traverse' (MSF msf) = MSF $ \xs -> do
+    fbm <- mapM msf xs
```
freckletonj, line 30: `fbm :: f (b, MSF m a b)`. So from that we need to return a tuple. The first half is easy and sensible. The second half confuses me. I would _like_ to use `fbm` to construct the 2nd half of the tuple, and `sequence` / `traverse` almost feel right, but, in the end I just reused the input parameter, which feels wrong, but type checks and seems to work. Thoughts?

turion: Yes, it can't be quite right because this way the internal state of the `MSF` will not advance. 

One strategy could be to recognise that the type of `unMSF :: MSF m a b -> a -> (b, MSF m a b)` is isomorphic to `unMSF' :: a -> StateT (MSF m a b) b`. Then the heartpiece is to do `traverse unMSF'`.

A similar approach is discussed here: https://github.com/turion/essence-of-live-coding/issues/57#issuecomment-876296306

turion: Let me know if you want more help with this :)

freckletonj: Great intuition, that made it quite simple.

I added a custom `StateT` and then tried to back it back out (to just an unwrapped fn), because I don't want to add an mtl dependency.

But alas, this instance doesn't exist:

```haskell
instanceish Functor m => Applicative (s -> m (a, s)) where  -- Applicative in that `a`
```

Since we can newtype a `StateT` equivalent, I'd recommend doing that instead of bringing in mtl, unless there's a way that the `traverse ...` can get at the correct `Applicative` instance? I'll think on this some more.

turion: You don't need to add `mtl` as a dependency. `StateT` is defined in `transformers`, and `dunai` already has a dependency on it. See https://hackage.haskell.org/package/transformers-0.5.6.2/docs/Control-Monad-Trans-State-Strict.html. It has all the instances you need.

I'm not sure whether we need strict or lazy state. I'm tending to strict state.

freckletonj: updated!

So I'm not sure I understand the tradeoff between strict and lazy state. I would guess that laziness might be necessary for ArrowLoop or MonadFix to work properly... But I'm not sure.

freckletonj:
updated!

So I'm not sure I understand the tradeoff between strict and lazy state. I would guess that laziness might be necessary for ArrowLoop or MonadFix to work properly... But I'm not sure.

turion:
The difference is whether in a state transition the internal state (i.e. the current MSF continuation) is evaluated to WHNF. In most cases this is done automatically on the next tick because it is needed then.

freckletonj:
@turion It's a humble PR, but it should be g2g now!
Thanks for your help writing it (the `StateT` thing was brilliant), and I'm happy to accept further advice.

dunai/src/Data/MonadicStreamFunction/Instances/Profunctor.hs
```diff
@@ -0,0 +1,45 @@
...
+  left' = left
+
+instance Monad m => Strong (MSF m) where
+  first' = first
+
+instance Monad m => Profunctor (MSF m) where
+  dimap l r p = arr l >>> p >>> arr r
+
+instance (Functor m, Applicative m, Monad m) => Traversing (MSF m) where
+  traverse' (MSF msf) =
```
turion, line 39:
```suggestion
  traverse' msf =
```

No need to unwrap here since you're wrapping again in line 44.

dunai/src/Data/MonadicStreamFunction/Instances/Profunctor.hs
```diff
@@ -0,0 +1,45 @@
...
+    let unMSF' x = StateT (\(MSF m) -> m x)
```
turion, line 40: You could also write `unMSF' = StateT . flip unMSF`. I'd personally prefer that style, and then inline it in line 44, but others might not.

dunai/src/Data/MonadicStreamFunction/Instances/Profunctor.hs
```diff
@@ -0,0 +1,45 @@
...
+instance Monad m => Choice (MSF m) where
+  left' = left
+
+instance Monad m => Strong (MSF m) where
+  first' = first
+
+instance Monad m => Profunctor (MSF m) where
+  dimap l r p = arr l >>> p >>> arr r
+
+instance (Functor m, Applicative m, Monad m) => Traversing (MSF m) where
```
turion, line 38: Maybe add a comment how the instance is defined, i.e. that the internal state of the MSF is stepped once for every element of the input `Traversable`

turion (changes requested):
Some small comments

turion:
You should definitely squash the commits ;)

freckletonj:
Good recommendations. And I squashed the commits, but if you haven't seen it yet, I tend to use github's `Merge Squash` feature that does that automatically once right at the end when you merge, it's kinda nice too.

Any other issues you'd like a hand with?

turion:
Looking great! @ivanperez-keera Is this fine to merge for you? I'm only asking since this introduces one dependency ([`profunctors`](https://hackage.haskell.org/package/profunctors)).

> Good recommendations. And I squashed the commits, but if you haven't seen it yet, I tend to use github's `Merge Squash` feature that does that automatically once right at the end when you merge, it's kinda nice too.

Right, I've just never used it. And I don't know what the commit message will be in that case.

> Any other issues you'd like a hand with?

Any that are open, really :)

turion (approved):

freckletonj:
It actually gives you a prompt to write a commit message.

Cool I'll take a look at open issues :)

ivanperez-keera:
I'll take a look and get back to you.

miguel-negrao:
I believe this would make `mapMaybeS` and `mapMSF` special cases of `traverse'`. Should also works for `Vector`, `Seq`, vector-sized's `Vector`, etc. I find it quite useful.

PR #260, opened 2021-10-10
Make afterEachCat behave like in yampa. Close #243

### LINKED ISSUE #243: Bearriver: afterEachCat doesn't behave like in Yampa

walseb:
I added the `Non-positive period` error because it's present in yampa.

GHC thinks `fireEvents` is a non-total function because it can recieve an empty list in place of `(qx:qxs)`. But that edge case is handled in `afterEachCat'`, which is the only function that calls it. Should we still l handle the edge case in `fireEvents` just to not have the the warning there?

All edge cases involving massive `dt`s should be handled. (for example with a `dt` of 1 and a time list [((0.5, "a"), (0.4, "b")]), both "a" and "b" should fire and the next event should be delayed by the remaining 0.1 seconds.

Thoughts on the multiline `if null ev`? I think it's slightly more readable than before but maybe that's because `null` being a value and not a function in most programming languages is throwing me off

### DIFF
--- a/dunai-frp-bearriver/src/FRP/BearRiver.hs
+++ b/dunai-frp-bearriver/src/FRP/BearRiver.hs
@@ -191,14 +191,25 @@ afterEachCat :: Monad m => [(Time,b)] -> SF m a (Event [b])
 afterEachCat = afterEachCat' 0
   where
     afterEachCat' :: Monad m => Time -> [(Time,b)] -> SF m a (Event [b])
-    afterEachCat' _ []  = never
+    afterEachCat' _ [] = never
     afterEachCat' t qxs = MSF $ \_ -> do
       dt <- ask
-      let t' = t + dt
-          (qxsNow, qxsLater) = span (\p -> fst p <= t') qxs
-          ev = if null qxsNow then NoEvent else Event (map snd qxsNow)
-      return (ev, afterEachCat' t' qxsLater)
+      let (ev, t', qxs') = fireEvents [] (t + dt) qxs
+          ev' = if null ev
+                  then NoEvent
+                  else Event (reverse ev)
 
+      return (ev', afterEachCat' t' qxs')
+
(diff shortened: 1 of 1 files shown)


ivanperez-keera:
Closed in favor of 54ae3be7cb2d0afc6dbbb9cafb4d632567929849.

See commit message of 9285e4edf22cec1176c7e881c64d38035f771802 for details on how it was verified. @walseb you have been credited as author of the commit.

Thanks!!

Issue #268, opened 2022-02-18
bouncingball - "drag" ?


miguel-negrao:
Letting the example "bouncingball" run for a while it behaves as if it had drag, the height that the ball reaches gets lower and lower until it doesn't move. And then it starts going down until disappears from the window. Both the Yampa and Bearriver versions behave the same. Is this due to loss of precision in the integration ? Is this a known issue ?

ivanperez-keera:
It is a know issue. You can check the numbers being produced, and yes, I think it's a rounding error.

miguel-negrao:
It takes more or less 1 minute until the issue arises. It feels a bit strange that one of the main examples of the library cannot run a simulation for more than a couple of seconds. I tried changing to Double but that didn't help. How do people creating games using the library deal with this issue ?

walseb:
I would add a conditional to stop it once its reached 0.

miguel-negrao:
Can you clarify what do you mean ? I believe the simulation is supposed to be a fully elastic collision, without losses of any kind, so it should go on jumping to the same initial height forever, right ?

walseb:
I see I thought the loss of force was intentional but not the negative force. Sorry for the confusion.

ivanperez-keera:
I think one way to go about it is to detect the actual point of collision. Right now, the point / time where the collision with 0 does happen is probably not being calculated properly, so the bounce is not fully elastic. There's a small error, which is smaller the faster you sample, but it's there.

miguel-negrao:
I see. The paper "Functional Reactive Programming, Refactored" touches on this and provides a solution with  `CCDGameMonad`. Then I would suggest changing this example either to the most simple version of collision detection that works correctly in Yampa/Bearriver, or change to something without collision such as a pendulum.

ivanperez-keera:
This example is useful because it's valid in both Dunai and Yampa. The example degenerates to the ideal meaning if the sampling rate is infinitely high. (I can also tell you that a game I built in the past used this, so it can be used --with small workarounds.)

Yampa cannot accept the custom monad, so that solution does not solve all problems.

But I still agree that it may be confusing, and it (rightfully) prompted this question. So perhaps 1) a comment in the file is in order, and 2) a small example with code that does not exhibit this bug could also be added separately.

miguel-negrao:
I gave it a go at estimating the velocity at the real point of impact (in between ticks) and then estimating the final velocity for that tick after the collision, but I wasn't successful.

[...]

PR #269, opened 2022-03-03
Fix BouncingBall.hs

miguel-negrao:
Fix an issue  (#268) with the physics simulation in BouncingBall.hs. Thansk to @ScottFreeCode for figuring out the problem.

### DIFF
--- a/dunai-examples/bouncingball/BouncingBall.hs
+++ b/dunai-examples/bouncingball/BouncingBall.hs
@@ -8,19 +8,30 @@ import FRP.Yampa     as Yampa
 #endif
 
 import Control.Concurrent
+import Data.IORef
+import System.Clock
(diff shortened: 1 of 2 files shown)


ivanperez-keera:
I'll probably also split this issue in several. My understanding is that the only change that is strictly needed to fix the issue is the use of `averageLast` (or something akin). Other changes, like the use of `System.Clock`, removing commented code, changes to the rendering code, etc. are separate. (They could still make sense separately, but they just have nothing to do with #268 if my understanding is correct so they should not be merged as part of this issue).

ivanperez-keera:
Closed in favor of 6fcf67628c24ba2608735938dcf7fa78c29ffa2b.

You (and others) have been credited for the fix.

Issue #245, opened 2021-03-20
MSF arrows aren't associative in terms of evaluation


walseb:
Hi! I was re-reading the old issue @turion helped me with (#236) and noticed that he mentioned the issue with MSF compose strictness way before I eventually figured out that was the problem, I think I just didn't make the connection. But more relevant to this issue, I was reminded of the problem I encountered there with MSF arrows not being associative. Back then I was more concerned with just getting my yampa stuff working in bearriver, but today I decided to give it a go.

I have managed to reduce it down to this:
``` haskell
main = reactimate (return ()) (\ _ -> return (0.1, Just ())) (\_ b -> print b >> pure False) sfB -- sfB

-- Runs
sfA :: SF () ()
sfA = constant undefined >>> identity >>> constant ()

-- Doesn't run
sfB :: SF () ()
sfB = (constant undefined >>> identity) >>> constant ()
```
In `sfA` the output of `constant undefined` is never evaluated as it isn't needed to create a result.
In `sfB` the output of `constant undefined` is reduced to WHNF which results in undefined being thrown.
But in both examples the output of `constant undefined` isn't actually needed to get the result so they shouldn't be evaluated even to WHNF. If we put the `undefined` into a data structure so it can't be fully evaluated with just WHNF we can see that this is indeed a evaluation problem:
``` haskell
-- Runs despite same parenthesis as sfB
sfC :: SF () ()
sfC = (constant (undefined, undefined) >>> identity) >>> constant ()
```

Beyond this I haven't made any progress yet. For easy reference here are the relevant functions:
``` haskell
constant = arr . const
identity = arr . id

f >>> g = g . f
sf2 . sf1 = MSF $ \a -> do
  (b, sf1') <- unMSF sf1 a
  (c, sf2') <- unMSF sf2 b
  let sf' = sf2' . sf1'
  c `seq` return (c, sf')
```

ivanperez-keera:
Good catch.

This problem seems to clearly stem from the use of `seq`.

Two observations:

- If needed I think we could provide a `seqA` if we don't have one already.
- I wonder if it's possible to always move the `seq` out of the definition of `(.)`. I think it is.

ivanperez-keera:
@walseb Can you please check what happens if you remove the `seq` from the definition of `(.)` and try again?

walseb:
[...]

PR #314, opened 2022-08-28
removed `seq` from "instance Category (MSF m)" (issue #245)

geroldmeisinger:
See issue #245. I needed a branch that we can refer to. Shouldn't probably be pushed to develop directly but I don't know how to pull request into a new branch. Please change manually.


ivanperez-keera:
Here: https://github.com/ivanperez-keera/dunai/tree/develop-no-seq

ivanperez-keera:
For now, the branch is created. We'll got to this either when we address #245, or when we address performance (which will happen when we deal with the tests in Yampa; see https://github.com/ivanperez-keera/Yampa/issues/217 and subissues for a summary).

PR #315, opened 2022-08-29
added Event type-class instances from Maybe

geroldmeisinger:
issue #327

### DIFF
--- a/dunai-frp-bearriver/src/FRP/BearRiver.hs
+++ b/dunai-frp-bearriver/src/FRP/BearRiver.hs
@@ -39,7 +39,9 @@ import           Data.Data
 import           Data.Functor.Classes
 import           Data.Functor.Identity
 import           Data.Maybe
+#if __GLASGOW_HASKELL__ >= 800
 import           Data.Semigroup
+#endif
 import           Data.Traversable          as T
 import           Data.VectorSpace          as X
 
--- a/dunai-frp-bearriver/src/FRP/BearRiver.hs
+++ b/dunai-frp-bearriver/src/FRP/BearRiver.hs
@@ -29,6 +29,7 @@ import           Control.Applicative
 import           Control.Arrow             as X
 import qualified Control.Category          as Category
 import           Control.Monad             (mapM)
+import           Control.Monad.Fail
 import           Control.Monad.Random
 import           Control.Monad.Trans.Maybe
 import           Control.Monad.Zip
--- a/dunai-frp-bearriver/src/FRP/BearRiver.hs
+++ b/dunai-frp-bearriver/src/FRP/BearRiver.hs
@@ -1,6 +1,8 @@
 {-# LANGUAGE Arrows     #-}
 {-# LANGUAGE CPP        #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE DeriveDataTypeable #-}
 -- The following warning is disabled so that we do not see warnings due to
 -- using ListT on an MSF to implement parallelism with broadcasting.
 #if __GLASGOW_HASKELL__ < 800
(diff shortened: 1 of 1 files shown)


ivanperez-keera:
Let's do this in order. What instances are you including, and are they missing in Yampa or only in bearriver?

geroldmeisinger:
All instances which are also defined for Maybe, see https://github.com/ivanperez-keera/dunai/pull/315/commits/030695062ed483b8b4f3afa401229c6ded8e6bed

I just work with BearRiver right now, don't know for Yampa.

ivanperez-keera:
Ok. Because bearriver implements Yampa, it makes sense to add things there first. Let's see what's missing in Yampa.

Also, are these "nice to have"s, or are they needed for something you are working on?

geroldmeisinger:
please consider all my pull request nice2have right now. They were easy to make and I just wanted to make them available for everyone to refer to. I need Foldable which relies on Semigroup but I'm working in my own fork so don't worry.

geroldmeisinger:
Yampa provides: Show (via deriving), Eq, Ord, Functor, Applicative, Monad, Fail.Monad.Fail, Alternative, NFData(?)

Why are you asking for Yampa anyway? May I ask how you see the relation between these projects? I was my understanding that:
* Yampa (old Yampa, Yampa 1) implemented SF directly and acts as a "standard" which shouldn't be changed fundamentally.
* Dunai (MSFs) provides Monadic Stream Functions (which are independent of time).
* BearRiver Yampa (new Yampa, Yampa 2) which tries to re-implement old Yampa based on Dunai's MSFs as close as possible. However [“There’s some fundamental differences, like the fact that, in principle, bearriver signals do not exist at time 0”
](https://discourse.haskell.org/t/dunai-bearriver-frp-new-version-and-hacktoberfest/909/3)

So why not provide more instances if it's a no-brainer?

ivanperez-keera:
Because it's never a no-brainer. It just looks like that at the beginning because the up-front cost of one different is small, but the long-term cost of many differences is big. It's a marathon, not a speed race. It's not about everything we can potentially add, it's about the value we'll obtain over time vs the cost of maintaining the code. The more disparate these are, the harder it becomes to maintain. When something's added to BR but not to Yampa, there has to be a very, very good reason (beyond "why not").

Bearriver is not Yampa 2. Yampa keeps being maintained and improved, and sees releases every 2 months.

Bearriver is an implementation of Yampa's API using MSFs. It is used 1) to show what MSFs can do, and 2) to extend Yampa where SFs alone can't reach.

When we get to performance evaluation of Yampa (that's in Yampa's roadmap), there is a chance that Yampa will be implemented using MSFs (if they are or can be made faster than Yampa's current implementation using SFs and GADTs).

At present, the flow of features between Yampa and Bearriver is:
- Features are added to Yampa (when it makes sense to add them).
- "Missing" features in Bearriver (those that Yampa has but BR doesn't) are added to BR.

Keeping BR as close to Yampa as reasonably possible greatly decreases the maintenance burden. We'll get to the same point, just in a more organized, predictable way, and without burning out (which is incredibly common).

ivanperez-keera:
>  missing in current BearRiver HEAD compared to Yampa: Ord, MonadFail, Alternative, NFData

I'd definitely accept these in BR.

ivanperez-keera:
> I need Foldable which relies on Semigroup

I'm curious about what you are doing and why `Foldable` was needed. It may point to a lack in Yampa that must be filled.

ivanperez-keera:
> Bearriver is not Yampa 2. Yampa keeps being maintained and improved, and sees releases every 2 months.

...and there's a reason why dunai and BR are released after Yampa every time (about 2 weeks after): so that the newest BR implements the newest Yampa.

geroldmeisinger:
> I'm curious about what you are doing and why Foldable was needed. It may point to a lack in Yampa that must be filled.

"if key is 'q' raise quit event Event () otherwise NoEvent"

```
let key = Event 'a'

let quitE0 = filterE (== 'q') key `tag`()
let quitE1 = mfilter (== 'q') key `tag` () -- requires MonadPlus
let quitE2 = boolToEvent ('q' `elem` key) -- requires Foldable
```

ivanperez-keera:
Why not just use:
```
let quitE0 = filterE (== 'q') key `tag`()
```

geroldmeisinger:
as of https://github.com/ivanperez-keera/dunai/issues/327 opened issue in https://github.com/ivanperez-keera/Yampa/issues/234 first

PR #316, opened 2022-08-29
added dpSwitch implementation by thalerjonathan (issue #40)

geroldmeisinger:
https://github.com/ivanperez-keera/dunai/issues/40#issuecomment-354439225

### DIFF
--- a/dunai-frp-bearriver/CHANGELOG
+++ b/dunai-frp-bearriver/CHANGELOG
@@ -1,3 +1,6 @@
+2022-09-06 Ivan Perez <ivan.perez@keera.co.uk>
+        * src/: added dpSwitch (#331).
+
 2022-08-21 Ivan Perez <ivan.perez@keera.co.uk>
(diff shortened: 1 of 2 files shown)


ivanperez-keera:
Is this tested? Does it behave the same way as Yampa's `dpSwitch`?

The same comments as in https://github.com/ivanperez-keera/dunai/pull/324#issuecomment-1236504476 would apply.

ivanperez-keera:
Plus squashing commits as appropriate.

ivanperez-keera:
Also, since this doesn't really close issue #40, it would create a commit tree in which it is unclear when issues are finished (all issues reference numbers, but a merge doesn't mean that the problem described is completely solved).

It's best to either create a separate issue for `dpSwitch`, or wait to add more definitions from Yampa and include it there.

ivanperez-keera:
This has now been made part of #426, which should be included in the next release.

PR #317, opened 2022-08-30
make haddock show sections, codedocs from Yampa

geroldmeisinger:
This makes the haddock documentation show contents, sections and structure again which I think is much more pleasing. Unfortunately it omits re-exports of other modules (Arrow, MSF, ...). It's arguable which version is a better, but I prefer this one because the links to other modules are still there.

I might not understand Haddock, anyway I filed a bug report here: https://github.com/haskell/haddock/issues/1524


geroldmeisinger:
image

image

PR #318, opened 2022-08-30
removed duplicate defintion of dup, already definied in utilities

geroldmeisinger:


ivanperez-keera:
I'm trying to be quite systematic with the way changes are documented and implemented. See specifically: https://github.com/ivanperez-keera/dunai#version-control

Let's open an issue for removing `dup`, follow the guide I linked to there, and ignore the extra space at the bottom for now. Removing the extra needed, but it's unrelated to the fact that a duplicated function can be removed, so it's a better practice to clean the file as part of a separate issue and not include it as part of this change.

ivanperez-keera:
I have created an issue #321 to document this. If you want, I'm happy to explain how the dunai process works so you can prepare the change yourself. That will also help include other changes you've proposed, and it'll make it much easier for me to review them and accept them.

ivanperez-keera:
If you can, please let me know by tomorrow. Otherwise, I'll finish it and merge the change. But it would be really great to get you up to speed with the process I'm following. That'll make adding future changes much easier.

ivanperez-keera:
Issue defined in #321 and closed by 4513307f971a0a2e715b13439797dad4929e4b56.

Issue #322, opened 2022-09-04
`bearriver`: haddock code docs are missing sections


geroldmeisinger:
At present, the module `FRP.Bearriver` does not show sections, which makes it much harder to navigate.

It looks like this:
![187398816-a66541ca-71dc-49d8-a40b-1c284438b7c2](https://user-images.githubusercontent.com/112266044/188312143-4b7dce24-554f-48b8-98e9-6e9558052ec4.png)

but, if we add sections, it would look like that:
![187398770-d9f82154-af5c-493d-af2e-d08e1b7b5a21](https://user-images.githubusercontent.com/112266044/188312141-6b21b4db-d5fa-4be4-96aa-4a068c63f1f1.png)

EDIT: Ivan Perez: Add a bit of wording to explain also in prose what needs to be done.

PR #324, opened 2022-09-04
`bearriver`: make haddock show sections. Refs #322.

geroldmeisinger:
This makes the haddock documentation show contents, sections and structure again which I think is much more pleasing. Unfortunately it omits re-exports of other modules (Arrow, MSF, ...). It's arguable which version is a better, but I prefer this one because the links to other modules are still there.

I might not understand Haddock, anyway I filed a bug report here: https://github.com/haskell/haddock/issues/1524

### DIFF
--- a/dunai-frp-bearriver/CHANGELOG
+++ b/dunai-frp-bearriver/CHANGELOG
@@ -1,3 +1,6 @@
+2022-09-05 Ivan Perez <ivan.perez@keera.co.uk>
+        * src/BearRiver.hs: show sections in haddock (#322).
+
 2022-08-21 Ivan Perez <ivan.perez@keera.co.uk>
         * bearriver.cabal: Version bump (0.13.6) (#302), add version bounds for
           dependencies (#303).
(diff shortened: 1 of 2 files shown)


ivanperez-keera:
Thanks.

Changes:
- Please modify the commit message so that it has the shape:
```
bearriver: <add your original subject line here>
```
- Add a period at the end of the subject line.
- Please add a bit of info to the body of the commit message. For a number of reasons, including that the issues on github may one day be lost, we want to make the history of the repo self-documenting. We often borrow from the issue description. The body of the commit message is also a good place to add details if the summary line of the commit becomes too long.
- Please put the options after the language pragmas (see our style guide, which lists pragmas first).
- Please also update the changelog in bearriver. Notice the way that the separate commit for the changelog is normally done (do a `git log` to see them).

ivanperez-keera:
The address I'm adding to the changelog is that of the maintainer. The best place to credit your work is with a Thanks @<username> under the issue list for each release.

See for example: https://github.com/ivanperez-keera/dunai/blob/develop/dunai/CHANGELOG#L47-L54 (that doesn't have the issue numbers next to the changes, but I'm only talking about the "Thanks:" line.

ivanperez-keera:
This has been merged. It had to be rebased. Thank you @geroldmeisinger .

PR #336, opened 2022-11-03
Introduce MSFAsync

turion:
Work in progress to clean up the module `Data.MonadicStreamFunction.Async`.

@ivanperez-keera would you in principle merge this? Then I would clean up, and document.

### DIFF
--- a/dunai/src/Data/MonadicStreamFunction/Async.hs
+++ b/dunai/src/Data/MonadicStreamFunction/Async.hs
@@ -38,13 +38,14 @@ instance Monad m => Monad (MSFAsync i m) where
             unMSFAsync $ f a
             go msf'
           Left () -> return ()
+
 -- |
 -- Concatenates a monadic stream of lists to a monadic stream.
 -- The stream of lists will be called exactly when new data is needed.
 --
 -- Example:
 --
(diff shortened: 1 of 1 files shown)


turion:
I got the basic insight from @AshleyYakeley in https://github.com/haskell/mtl/issues/85#issuecomment-1298231912. Since `MSF (MaybeT m) a b == ListT (ReaderT a m) b`, it should have a monad instance. This PR implements it, and gets the function `concatS` for free.

ivanperez-keera:
A couple of comments:
- Can you first create an issue to document what's happening?
- Don't mix fixes to other things as part of the same PR (even if they are in the same module).
- The new type should be by default hidden in the module. It's an implementation artifact.
- How does the new implementation of concatS work? It looks pretty complicated to me. Can you document it and potentially introduce intermediate names (and types) for the subexpressions?

ivanperez-keera:
BTW we also have a discussion tab were we can start the conversation. One things are clearer we can turn it into an issue. That can help separate long conversations (_discussions_) from more formal (and succinct) change proposals (_issues_).

turion:
> Can you first create an issue to document what's happening?

This is https://github.com/ivanperez-keera/dunai/issues/338

Issue #438, opened 2024-10-12
`bearriver`: `loopPre` is defined twice


ivanperez-keera:
The function `loopPre` is defined twice: once in `FRP.BearRiver`, and once in `FRP.BearRiver.Loop`. Only one definition should be needed.

This prevents `FRP.BearRiver` from providing an API equivalent to Yampa's.

ivanperez-keera:
My intuition here would be to just remove the one in `FRP.BearRiver`.

pbrinkmeier:
Hi, I'd like to work on this issue. The second approach seems less likely to break the API, so I would perhaps go with that?

ivanperez-keera:
Great!

Yeah, let's go with the second approach. Can you please structure it this way?

- Commit 1: Expose FRP.Yampa.Loop in FRP.Yampa hiding FRP.BearRiver.loopPre
- Commit 2: Document changes in CHANGELOG.

Always reference the issue with `. Refs #438.` at the end of the commit message summary, etc.

I need to have the next release ready tonight (EOB Pacific time). Do you think you can have it ready for then?

pbrinkmeier:
I applied the changes in #441. The duplicate definition of `loopPre` is removed and replaced by re-exports where required. Please check if I understood the task, then I will add a changelog entry.

ivanperez-keera:
Thanks! I responded in the PR.

PR #441, opened 2024-10-20
bearriver: Remove duplicate loopPre definition. Refs #438.

pbrinkmeier:
Removes the loopPre definition from FRP.BearRiver and replaces it by a re-export from FRP.Yampa.Loop.

### DIFF
--- a/dunai-frp-bearriver/CHANGELOG
+++ b/dunai-frp-bearriver/CHANGELOG
@@ -1,7 +1,8 @@
-2024-10-19 Ivan Perez <ivan.perez@keera.co.uk>
+2024-10-21 Ivan Perez <ivan.perez@keera.co.uk>
         * Offer all definitions from FRP.Yampa.Switches (#426).
         * Publish embed specialized for FRP.Yampa's SF (#439).
         * Increase upper bounds on MonadRandom (#436).
+        * Re-export FRP.BearRiver.Loop in FRP.Yampa (#438).
 
 2024-08-21 Ivan Perez <ivan.perez@keera.co.uk>
         * Version bump (0.14.10) (#430).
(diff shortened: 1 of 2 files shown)


ivanperez-keera:
So, what I had in mind was:
- leaving `FRP.BearRiver` as is
- in `FRP.Yampa`
  - hiding `loopPre` in the import of `FRP.BearRiver` on line 12: https://github.com/ivanperez-keera/dunai/blob/6aefd90e941493fbf6168e29df1fdd8ab324b42b/dunai-frp-bearriver/src/FRP/Yampa.hs#L12
  - additionally importing `FRP.BearRiver.Loop as X`, and re-exporting it from `FRP.Yampa` (module X is already exported so it's all good there).

pbrinkmeier:
I see! I assumed the goal was to remove one of the duplicate `loopPre` definitions. I changed the code according to your comment.

ivanperez-keera:
There may have been some confusion. What I understood in "https://github.com/ivanperez-keera/dunai/issues/438#issuecomment-2408658014" with "the second approach" was "hide it in `FRP.Yampa` and re-export `FRP.BearRiver.Loop` from `FRP.Yampa`.".

This looks good. Can you add the changelog entry as a separate commit?

pbrinkmeier:
Done!

ivanperez-keera:
@pbrinkmeier  Thanks again for your contribution!

Thanks also for creating a branch dedicated for this change with a specific name `develop-remove-looppre-duplication`. That was a nice detail :)

Just FYI, I made a couple of changes. One was aligning the imports, for which I just piped the imports through `stylish-haskell` (my mistake for not saying anything earlier). The other was that I modified the commit message to be more specific to the new change, and give a tiny bit more detail for future maintainers.

I wish everyone made it as easy as you to integrate their changes!

Issue #452, opened 2024-11-03
`dunai-examples`: Example re-defines existing function in `FRP.Yampa`


ivanperez-keera:
The example `bearriver-examples-bouncingball-list` defines a function `arr2` equivalent to Yampa's original `arr2` function.

The fix for #360 defined all functions offered by the corresponding Yampa module, so that re-definition as part of the example is no longer necessary.

The example no longer compiles for that reason.

ivanperez-keera:
If somebody would like to work on this, please send a PR that does the following:

- Commit 1: Remove these lines:
https://github.com/ivanperez-keera/dunai/blob/04c1dc3e9648b9f43c1ee55855ed005e0fb9d80b/dunai-examples/list/BouncingBall.hs#L119-L121

- Set as commit message:
```
dunai-examples: Remove duplicated function. Refs #452.

The example bearriver-examples-bouncingball-list defines a function arr2 equivalent to Yampa's original arr2 function. The fix for issue #360 defined all functions offered by the corresponding Yampa module, so that re-definition as part of the example is no longer necessary. The example no longer compiles for that reason.

This commit removes arr2, so that the function coming from `FRP.Yampa` is used instead.
```

Issue #360, opened 2023-04-24
`bearriver`: Offer all definitions from `FRP.Yampa.Arrow`


ivanperez-keera:
The module `FRP.Yampa.Arrow` is being offered by Yampa, but not `bearriver`.

This makes the interface offered by `bearriver` not compliant with `Yampa`, therefore not delivering on the promise of being API-compatible.

This is a subissue of #40. Because #40 is a big issue, it's good to split it in several smaller ones.

PR #453, opened 2024-11-03
dunai-examples: Remove duplicated function. Refs #452.

solomon-b:
The example bearriver-examples-bouncingball-list defines a function arr2 equivalent to Yampa's original arr2 function. The fix for issue #360 defined all functions offered by the corresponding Yampa module, so that re-definition as part of the example is no longer necessary. The example no longer compiles for that reason.

This commit removes arr2, so that the function coming from `FRP.Yampa` is used instead.


solomon-b:
NOTE: Without fixing `reactimate` the example will still not build successfully.

ivanperez-keera:
Thanks again, @solomon-b . Merged!

Issue #457, opened 2024-11-09
`dunai-example`: Example hangs on mouse click


ivanperez-keera:
The example `dunai-examples/list/` is supposed to demonstrate how new elements can be added dynamically in a dunai animation by leveraging the list monad, by allowing users to create new elements with a mouse click.

However, it currently hangs when one clicks on the screen.

This is a problem with a switch switching repeatedly.

ivanperez-keera:
@mchav I'v created this separate issue to fix this specific bug, since it's technically unrelated to the change to SDL2 and to the issue with reactimate. Would you like to send a PR?  If so, could you please make the first commit fix the issue, and then add another commit that updates the changelog?

Feel free to ask me or check other issues and the history for details on how to document commits, commit messages, and the changelog.

PR #458, opened 2024-11-10
dunai-examples: Fix bug where example hangs on click. Refs #457.

mchav:
Switch is currently stuck in an infinite loop.
`notYet` delays the switch event to the next point in time T(n + 1). The switch will now happen AFTER the event at T(n) has passed.

This issue is described in the [Yampa Arcade
paper](https://dl.acm.org/doi/10.1145/871895.871897) on page 8.

> The resulting signal function is composed with
> `notYet :: SF (Event a) (Event a)`
> that suppresses initial event occurrences.
> Thus the overall result is a source of kill and
> spawn events that will not have
> any occurrence at the point in time when it is
> first activated. This is to prevent gameCore
> from getting stuck in an infinite loop of
> switching.

This does change the semantics of the program
though since the switching now takes place
at T(n + 1) so the program is a little faster.

### DIFF
--- a/dunai-examples/list/BouncingBall.hs
+++ b/dunai-examples/list/BouncingBall.hs
@@ -43,6 +43,7 @@ fireballs :: SF (Bool, (Float, Float)) [(Float, Float)]
 fireballs = switch
   (    arr (const [])
    &&& arr (\(mp, pos) -> if mp then Event pos else Yampa.NoEvent)
+   >>> second notYet
   )
 
   (\(p, v) -> let oldfb = voidI $ runListMSF (liftTransS (bouncingBall p v))


ivanperez-keera:
@mchav Thanks a lot for your contribution. I made a couple of changes:

- I removed the entry from the `cabal.project`, because I don't want to commit that change. That file is used by the CI server and I don't want the CI job to have to do that right now. (You can still keep that change locally as you work though.)

- I kept the type signatures in. Your original PR removed several type signatures, but there's not reason to remove them. The code works and compiles just fine with them in (the types have not changed). In general, I think it's best to add type signatures because they help the developer understand the code.

- I modified the commit message slightly to add a bit more context, so that anyone looking at the history doesn't have to come to the issue on github to understand what the problem was about.

I recommend you sync your `develop` branch with the new `develop`.

